PR #1, opened 2020-06-23
Fix the maven wrapper, most of which was missing 

sambsnyd:
This was due to an erroneous .gitignore entry


jkschneider:
Thanks!

PR #2, opened 2020-07-11
enable GitHub Actions CI

sullis:


jkschneider:
First experience with Github Actions CI, let's see what this does! Thanks @sullis .

PR #56, opened 2020-10-20
instructions to run 'git apply' contain non-existing flag

mattnworb:
There does not appear to be a `-f` flag for `git apply`:

https://git-scm.com/docs/git-apply


jkschneider:
Thank you @mattnwob

Issue #59, opened 2020-10-23
rewrite:publish fails for projects with parents that are not in Maven Central


mattnworb:
I'm not sure if this is meant to be supported, but I noticed that the `publish` mojo fails if the project has a parent which is not found in Maven Central (in this case, the parent is deployed in an internal Maven repository):

```
[INFO] --- rewrite-maven-plugin:2.1.1:publish (default-cli) @ rewriter-mockito-example ---
[ERROR] unable to resolve model
org.apache.maven.model.resolution.UnresolvableModelException: Could not find artifact com.spotify:root:pom:64 in central (https://repo1.maven.org/maven2/)
    at org.apache.maven.project.ProjectModelResolver.resolveModel (ProjectModelResolver.java:196)
    at org.openrewrite.maven.internal.ParentModelResolver.resolveModel (ParentModelResolver.java:82)
    at org.openrewrite.maven.internal.ParentModelResolver.resolveModel (ParentModelResolver.java:92)
    at org.apache.maven.model.building.DefaultModelBuilder.readParentExternally (DefaultModelBuilder.java:1070)
    at org.apache.maven.model.building.DefaultModelBuilder.readParent (DefaultModelBuilder.java:846)
    at org.apache.maven.model.building.DefaultModelBuilder.build (DefaultModelBuilder.java:337)
    at org.openrewrite.maven.MavenModuleLoader.load (MavenModuleLoader.java:147)
    at org.openrewrite.maven.MavenModuleLoader.load (MavenModuleLoader.java:90)
    at org.openrewrite.maven.MavenModuleLoader.load (MavenModuleLoader.java:82)
    at org.openrewrite.maven.MavenParser.parseInputs (MavenParser.java:62)
    at org.openrewrite.Parser.parse (Parser.java:38)
    at org.openrewrite.maven.RewritePublishMojo.buildCycloneDxBom (RewritePublishMojo.java:107)
    at org.openrewrite.maven.RewritePublishMojo.execute (RewritePublishMojo.java:51)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:210)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
[...]

PR #63, opened 2020-11-06
publish: add 'skipCycloneDxBom' parameter

mattnworb:
This change allows the publish mojo to skip the building and attaching
of a CycloneDX BOM artifact to the project if the parameter
`skipCycloneDxBom` is set to true (default value is false).

Related to [#59][0], generating this artifact fails for projects where
the parent pom is not in Maven Central, so adding this property is
mostly a workaround to the failure but might have other uses as well.

[0]: https://github.com/openrewrite/rewrite-maven-plugin/issues/59


jkschneider:
Thanks @mattnworb! Also appreciate the careful work on the integration test.

Issue #79, opened 2021-01-19
Automate publishing of the plugin to maven central


tkvangorder:
Currently, all projects other than the `rewrite-maven-plugin` can be published via an automated ci/cd step. However, this project uses maven and GitHub actions and does not have the ability to automatically publish to maven central. 

This article provides a good overview of how this can be accomplished: 

https://dzone.com/articles/how-to-publish-artifacts-to-maven-central

aegershman:
No guarantees this link will exist as a repo in the future, but, testing this out over in a personal org/repo to not pollute publishes (https://github.com/slugstack/maven-release-example/actions?query=workflow%3Arelease)

Notes for reference:

- http://maven.apache.org/maven-release/maven-release-plugin/prepare-mojo.html
- https://docs.github.com/en/actions/guides/publishing-java-packages-with-maven
- for scm urls, https://maven.apache.org/scm/git.html -- keep in mind that `https` scm prefixes will play nicely with using github tokens for committing and pushing back, whereas most manual flows are expecting git/ssh for using people's stored ssh keys

aegershman:
used a throwaway repo for messy testing (https://github.com/slugstack/maven-release-example). Will have a PR for this squared up soon

aegershman:
Just as an update, I'm generally mistrustful of processes like this unless demonstrated and tested all the way through, so I'm testing this out against a (personal) maven central repo to test out the full deployment process before committing: https://github.com/slugstack/rewrite-definitions/runs/1744643853?check_suite_focus=true 

It's been worth having it barf up once or twice mid-release to see what to do from the perspective of "the release info has already been updated in git, but it's not reflected on external sources--now what?". Still finishing this up.

aegershman:
It's a little annoying that the release plugin commits with the 'new' version, _then_ pushes those to source, _then_ attempts pushing to repos after that, all in one go. It's kind of annoying that a rollback also involves a commit forward (because it required a commit in the first place). In reality what's really wanted is: "`tag` this commit with a version. then, take source of whatever is at that tagged version and distribute it. track where it was distributed to and if it was successful. if distributing it fails, mark it as failed so we can try again later. if it succeeds, mark that it's been successfully deployed." ... I suppose solving this in the general case is a bit more involved than what we want to do here. It's just an annoying category of problem altogether. I'll poke more to get it simplified. Just making notes here.

aegershman:
[...]

PR #81, opened 2021-01-22
github actions for automated release workflow

aegershman:
part of https://github.com/openrewrite/rewrite-maven-plugin/issues/79

This demonstrates how the current workflow of releasing could be uplifted to github actions; the biggest benefit being not requiring a shared set of credentials directly on the releaser's computers -- instead, stored as repository secrets.

Currently can be ran manually ran from the `actions` UI (though would be beneficial to one day have this trigger automatically on a `git tag` and push rather than the other way around).

An example of this in action can be seen here: https://github.com/slugstack/rewrite-definitions/runs/1751552089?check_suite_focus=true

Requires the following github secrets defined on the repository:

MAVEN_GPG_PASSPHRASE
MAVEN_GPG_PRIVATE_KEY
OSSRH_TOKEN
OSSRH_USERNAME

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,8 +33,8 @@ jobs:
           git config user.name "github-actions[bot]"
 
       - name: release
-        timeout-minutes: 30
-        run: ./mvnw --settings=${{ github.workspace }}/settings.xml --file=pom.xml -Prelease --batch-mode release:prepare release:perform
+        timeout-minutes: 45
+        run: ./mvnw --settings=${{ github.workspace }}/settings.xml --file=pom.xml --activate-profiles=release,release-automation --batch-mode help:active-profiles release:prepare release:perform
         env:
           GITHUB_TOKEN: ${{ github.token }}
           MAVEN_USERNAME: ${{ secrets.OSSRH_USERNAME }}
@@ -43,6 +43,6 @@ jobs:
 
       - name: rollback
         if: ${{ failure() }}
-        run: ./mvnw --settings=${{github.workspace}}/settings.xml --file=pom.xml -Prelease --batch-mode release:rollback
+        run: ./mvnw --settings=${{github.workspace}}/settings.xml --file=pom.xml --activate-profiles=release,release-automation --batch-mode help:active-profiles release:rollback
         env:
           GITHUB_TOKEN: ${{ github.token }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,48 @@
+---
+name: release
+
+on:
+  workflow_dispatch: {}
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: set-up-jdk
+        uses: actions/setup-java@v1
+        with:
+          java-version: "11"
+          server-id: ossrh
+          settings-path: ${{ github.workspace }}
+          server-username: MAVEN_USERNAME
+          server-password: MAVEN_PASSWORD
+          gpg-private-key: ${{ secrets.MAVEN_GPG_PRIVATE_KEY }}
+          gpg-passphrase: MAVEN_GPG_PASSPHRASE
+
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -34,9 +34,9 @@
 
     <scm>
         <connection>scm:git:https://github.com/openrewrite/rewrite-maven-plugin.git</connection>
-        <developerConnection>scm:git:ssh://git@github.com/openrewrite/rewrite-maven-plugin.git</developerConnection>
+        <developerConnection>scm:git:https://github.com/openrewrite/rewrite-maven-plugin.git</developerConnection>
```
aegershman, line 37: this is to allow for a `git push` to occur using a `{{github.token}}`. A token works with https, whereas `ssh` requires an actual ssh key. The tradeoff is we get a `github.token` for free in github actions, but if a _human_ were to try to manually do this on their local computer, they'd need to have a github token for auth rather than using what I think most people use-- an ssh key. It's not a huge deal, it's just something to note in case there's a need for manual intervention. 

I could see if the `scm` can be modifiable under a `profile` to use `https` for `release` and `ssh` for everything else

.github/workflows/release.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          path: ~/.m2/repository
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
+
+      - name: configure-git-user
+        run: |
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config user.name "github-actions[bot]"
```
sambsnyd, line 33: Perhaps now is the time to create a bot account for our organization(s). You should have all the permissions you need to create one. Maybe ask in #general on slack for people to suggest names.

.github/workflows/release.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: set-up-jdk
+        uses: actions/setup-java@v1
+        with:
+          java-version: "11"
+          server-id: ossrh
+          settings-path: ${{ github.workspace }}
+          server-username: MAVEN_USERNAME
+          server-password: MAVEN_PASSWORD
+          gpg-private-key: ${{ secrets.MAVEN_GPG_PRIVATE_KEY }}
+          gpg-passphrase: MAVEN_GPG_PASSPHRASE
```
sambsnyd, line 21: Should the username, password, and gpg_passphrase also be coming from `secrets`?

aegershman: They are, they're loaded in through `env` vars in the next step. This is setting up the env variable name for resolution resolution in `settings.xml` in the next step below. They're just named a little different, though it can change. The names are setup like this to show the difference between how the `settings.xml` var of `MAVEN_USERNAME` is setup vs. how it's resolved against the `env` var of `OSSRH_USERNAME` passed in through `secrets` in the next step 👍 

I'm under the impression it's setup this way to reduce risks with credentials being persisted to disk on the action runner--  see the `setup-java` example for info: https://github.com/actions/setup-java#publishing-using-apache-maven

.github/workflows/release.yml
aegershman: The only thing I don't have is the ability to create `@moderne.io` emails, which would be necessary for creating a `github` bot account with a proper git config username. Case-in-point, I created a github bot account using my `aaron@moderne.io` email here (https://github.com/openrewrite-ci-bot), but it wouldn't have it's own proper email association to it.

The git config is really just for display; it may be easier to just keep the `github-actions[bot]` username for now. Alternatively, we could put down `openrewrite` as the username, and `jkschneider@gmail.com` or `jonathan@moderne.io` as the email, and since we're using jonathan's `gpg` key, it'll show up as `Jonathan` as the committer 🤷

sambsnyd: github-actions is fine for now. We can create a service account later

.github/workflows/release.yml
sambsnyd: Got it thanks for explaining

PR #108, opened 2021-03-27
ci: add JDK 15, JDK 16

sullis:


jkschneider:
I would expect these to fail right now because we don't yet have a Java 15/16 Java parser.

sambsnyd:
We will be happy to merge this once we're able to support it, but aren't sure exactly when that will be

Issue #138, opened 2021-04-21
General (but specifically, discovery) UI usability enhancements


aegershman:
- Better "help" discovery
- Selectively being able to display further information such as description, sub-recipes, options.

```sh
./mvnw rewrite:discover -Dinteractive
```

https://user-images.githubusercontent.com/5619476/117037904-a49d2500-accc-11eb-97fb-ccb4ae4948e4.mov

PR #150, opened 2021-04-23
recipe configuration reference information for discover

aegershman:
see https://github.com/openrewrite/rewrite-maven-plugin/issues/138
moved backwards-compatible refactoring work to https://github.com/openrewrite/rewrite-maven-plugin/pull/151 and merged

- Outputs activeStyle configuration
- configurable maximum `recipeList` recursion using `-Drecursion=<num>`
- configurable level of detail to show similar to `:help` using `-Ddetail=<true|false>`.
- Move "active" recipes to the bottom of the UI output so people don't have to scroll back up top to see what's registered as active
- Summary moved to the bottom of the screen output

It feels like it may be more appropriate to have "active configuration" displayed all the time for every command rather than just discover. Discover may not be the right word.

default output

Screen Shot 2021-04-24 at 5 01 01 PM

Screen Shot 2021-04-25 at 5 55 13 PM


sambsnyd:
Looks good so far. I think we could probably drop the "name:" prefix from most lines in the log output

aegershman:
I hear ya; I'm going to take a bit to re-think some things with the default verbosity and display. Marked back as draft*

PR #156, opened 2021-04-30
Add logging when running a recipe

rbosse:
I added some logging to know where we are with the processing. 

Thanks!


aegershman:
Hey @rbosse right on, thanks. Out of curiosity-- from your perspective, as a user, would you expect these details to always be displayed on by default? Or behind some flag like `verbose=true`?

aegershman:
Sounds good, I'll get back to you on this soon 👍

PR #201, opened 2021-08-07
Added classpath to JavaProvenance.

traceyyoshima:


src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
```diff
@@ -296,45 +296,61 @@ protected ResultsContainer listResults() throws MojoExecutionException {
...
                     .styles(styles)
-                    .classpath(
-                            Stream.concat(
-                                    project.getCompileClasspathElements().stream(),
-                                    project.getTestClasspathElements().stream()
-                            )
-                                    .distinct()
-                                    .map(Paths::get)
-                                    .collect(toList())
-                    )
```
sambsnyd, line 307: This isn't right. You got rid of the classpath. The parser wont be able to type attribute anything.
What was there before also wasn't quite right, as it mixed the compile and test classpaths together.

The correct thing to do is have one `JavaParser` for "compile" scope sources and another for "test" scoped sources, with appropriate classpaths.

sambsnyd (changes requested):

src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
sambsnyd: Nevermind, I see where you do this later

PR #248, opened 2021-11-21
Add integration tests for init and configure mojos

timtebeek:
Was playing around with these locally, and noticed they each take a differently formatted `dependencies` argument, which isn't immediately clear from the outset. It's due to on the one hand `AddPlugin` in `init`, versus `ChangePluginDependencies` in `configure`. Having tests that verify and highlight these differences should help discoverability.

I'd even be inclined to change most likely the `init` mojo to use the gav form of dependencies used in `configure`. I guess that could be achieved by also using `.doNext(new ChangePluginDependencies(groupId, artifactId, dependencies))` in `init` instead of adding the dependencies through `AddPlugin`. What would you say to such a suggested change?

Also added a log statement as inexplicably (for me) the InitMojoIT silently ignored the mojo run, leaving me puzzled.
And I delay the maven source and javadoc plugins to be able to run `mvn test -Dtest=InitMojoIT` quicker locally.
Let me know if either of these changes should be reverted! :)

### DIFF
--- a/src/test/java/org/openrewrite/maven/InitMojoIT.java
+++ b/src/test/java/org/openrewrite/maven/InitMojoIT.java
@@ -1,0 +1,32 @@
+package org.openrewrite.maven;
+
+import com.soebes.itf.jupiter.extension.MavenGoal;
+import com.soebes.itf.jupiter.extension.MavenJupiterExtension;
+import com.soebes.itf.jupiter.extension.MavenTest;
+import com.soebes.itf.jupiter.extension.SystemProperties;
+import com.soebes.itf.jupiter.extension.SystemProperty;
+import com.soebes.itf.jupiter.maven.MavenExecutionResult;
+
+import static com.soebes.itf.extension.assertj.MavenITAssertions.assertThat;
+
+@MavenJupiterExtension
+@MavenGoal("${project.groupId}:${project.artifactId}:${project.version}:init")
+@SystemProperties({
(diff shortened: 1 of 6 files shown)


src/test/java/org/openrewrite/maven/InitMojoIT.java
```diff
@@ -0,0 +1,32 @@
...
+import com.soebes.itf.jupiter.maven.MavenExecutionResult;
+
+import static com.soebes.itf.extension.assertj.MavenITAssertions.assertThat;
+
+@MavenJupiterExtension
+@MavenGoal("${project.groupId}:${project.artifactId}:${project.version}:init")
+@SystemProperties({
+        @SystemProperty(value = "activeRecipes", content = "org.openrewrite.java.testing.junit5.ParameterizedRunnerToParameterized"),
+        @SystemProperty(value = "dependencies", content = "<dependencies><dependency><groupId>org.openrewrite.recipe</groupId><artifactId>rewrite-spring</artifactId><version>4.14.1</version></dependency></dependencies>"),
+        @SystemProperty(value = "rootOnly", content = "false")
```
timtebeek, line 17: There seems to be a mismatch in the first conditional in InitMojo, which is why this rootOnly=false is needed; didn't know if there was another proper fix, but this works, and InitMojo works in practice as well.

timtebeek:
Made a separate MR for my suggestion of [adopting GAV for dependencies in InitMojo](https://github.com/timtebeek/rewrite-maven-plugin/pull/1) as well; seems much cleaner than xml. Let me know if I should merge/retarget that one!

Issue #152, opened 2021-04-25
Use maven-provided Settings resolution for mavenParser


aegershman:
Noticed this as part of work in https://github.com/openrewrite/rewrite-maven-plugin/issues/138

Logging this for discussion before getting to it, though it's a straightforward change

We should have Maven inject the `settings` object for our `mavenParserBuilder.activeProfiles(...)` to use rather than trying to resolve the `settings.xml` ourselves. This will make things easier for end-users expecting to be able to set the settings.xml to a different location.

Referring to these lines: https://github.com/openrewrite/rewrite-maven-plugin/blob/d54cc97f9bf4c6911603d54ec2d0f5b7491631e6/src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java#L147-L162

There's two reasons to avoid trying to resolve the location of `~/.m2/settings.xml` ourselves:
- 1.) there are global-provided settings.xml, and user-provided settings.xml locations, both of which are merged by maven to be the final effective settings.xml after resolution. 
- 2.) if the user passes a custom location for the settings.xml by doing `--settings=some/path/to/settings.xml`. This might seem weird at first, but it's done quite often especially in ci/cd systems.

Instead we can just inject the Maven settings object like so

```java
    /**
     * The Maven settings.
     */
    @Parameter(defaultValue = "${settings}", readonly = true, required = true)
    private Settings settings;
```

and then reference the resolved `settings` object as we are currently doing.

See: http://maven.apache.org/plugin-tools/maven-plugin-tools-annotations/
See: https://github.com/khmarbaise/maven-it-extension/discussions/193#discussioncomment-636982

aegershman:
Cross-referencing links:
- https://github.com/openrewrite/rewrite-maven-plugin/issues/256
- https://github.com/ruben-garciapariente/test-openwrite/tree/maven-system-property

/cc @ruben-garciapariente By the way, will ensure resolving this covers the scope of https://github.com/openrewrite/rewrite-maven-plugin/issues/256 👍

aegershman:
@ruben-garciapariente This is fixed as of the latest snapshot on this repository, 4.17.0-SNAPSHOT. Thanks again for your sample repository, it helped troubleshoot and validate this easier.

If you get the opportunity to try this out using the latest snapshot of the maven plugin, let me know if something doesn't look right. Otherwise, we should be good to go. Thanks again 👍

jkschneider:
[...]

PR #262, opened 2021-12-07
Let Maven resolve settings.xml content for us, get values from MavenExecutionRequest

aegershman:
See: https://github.com/openrewrite/rewrite-maven-plugin/issues/152


sambsnyd (approved):
Looks reasonable

PR #267, opened 2021-12-15
Settings decryption request

aegershman:
Taking inspiration from https://github.com/apache/maven-plugins/blob/a007e769ed5825774d5c31ec06c0013c8ee2b4d4/maven-scm-publish-plugin/src/main/java/org/apache/maven/plugins/scmpublish/AbstractScmPublishMojo.java#L317-L336

https://github.com/openrewrite/rewrite-maven-plugin/issues/268

Something to be aware of: is there a possible `null` on getServers() if decryption fails?


rdifrango:
@aegershman this largely looks correct, just need to be sure to auth correctly to the server :)

PR #285, opened 2022-02-02
refactor: Migrate Java 8 to Java 11

sjungling:


murdos:
This highlights an issue I also encountered with this recipe: comments are removed from pom.xml.
I did not found time yet to investigate and identify the sub-recipe that causes that.

PR #312, opened 2022-03-17
Removed unused declarations of styles in MavenMojoProjectParser

thomaszub:
Hi,

I noticed `MavenMojoProjectParser` has unused declarations of styles. As `NamedStyles.merge` has no side effects, they should be safely removable.

Kind regards
Thomas


jkschneider:
Thanks!

PR #348, opened 2022-05-03
Added quark parser to `ResourceParser`

### LINKED ISSUE #346: Add `Quark` support to scoop up any missed files within a Maven project.
See https://github.com/openrewrite/rewrite/pull/1737

We need to add the QuarkParser as the last parser in the resource parser. This will pick up any misc. files that have been missed by  Rewrite's other parsers. Additionally, by putting this in the ResourceParse, any files within the main/test folders will have the proper `SourceSet` assigned to them.

traceyyoshima:
Changes:

- Files ignored due to threshold are parsed as a `Quark`.
- Added `QuarkParser` to `ResourceParser`.
- ResourceParser will only parse files in the current maven project.
- Replaced `Files.find` with `Files.walkFileTree`.

fixes #346

### DIFF
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -344,10 +344,9 @@ public class MavenMojoProjectParser {
 
         ResourceParser rp = new ResourceParser(logger, exclusions, sizeThresholdMb);
 
-        Set<Path> quarks = new HashSet<>();
         // Any resources parsed from "main/resources" should also have the main source set added to them.
         sourceFiles.addAll(ListUtils.map(
-                rp.parse(baseDir, mavenProject.getBasedir().toPath().resolve("src/main/resources"), alreadyParsed, quarks),
+                rp.parse(baseDir, mavenProject.getBasedir().toPath().resolve("src/main/resources"), alreadyParsed),
                 addProvenance(baseDir, ListUtils.concat(projectProvenance, javaParser.getSourceSet(ctx)), null)));
 
         logger.info("Parsing Java test files...");
@@ -367,12 +366,12 @@ public class MavenMojoProjectParser {
 
         // Any resources parsed from "test/resources" should also have the test source set added to them.
         sourceFiles.addAll(ListUtils.map(
-                rp.parse(baseDir, mavenProject.getBasedir().toPath().resolve("src/test/resources"), alreadyParsed, quarks),
+                rp.parse(baseDir, mavenProject.getBasedir().toPath().resolve("src/test/resources"), alreadyParsed),
                 addProvenance(baseDir, ListUtils.concat(projectProvenance, javaParser.getSourceSet(ctx)), null)));
 
         // Parse non-java, non-resource files
         sourceFiles.addAll(ListUtils.map(
-                rp.parse(baseDir, mavenProject.getBasedir().toPath(), alreadyParsed, quarks),
+                rp.parse(baseDir, mavenProject.getBasedir().toPath(), alreadyParsed),
                 addProvenance(baseDir, projectProvenance, null)
         ));
 
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -33,13 +33,11 @@ import org.openrewrite.maven.cache.MavenPomCache;
 import org.openrewrite.maven.cache.RocksdbMavenPomCache;
 import org.openrewrite.maven.internal.RawRepositories;
 import org.openrewrite.maven.tree.ProfileActivation;
-import org.openrewrite.quark.QuarkParser;
 import org.openrewrite.style.NamedStyles;
 import org.openrewrite.xml.tree.Xml;
 
 import java.io.File;
 import java.io.IOException;
-import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -381,7 +381,7 @@ public class MavenMojoProjectParser {
 
         // Parse non-java, non-resource files
         sourceFiles.addAll(ListUtils.map(
-                rp.parse(mavenSession.getCurrentProject().getBasedir().toPath(), alreadyParsed),
+                rp.parse(mavenProject.getBasedir().toPath(), alreadyParsed),
                 addProvenance(baseDir, projectProvenance, null)
         ));
 
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -306,10 +306,16 @@ public class MavenMojoProjectParser {
         return rawRepositories;
     }
 
+    private Set<Path> skipOtherMavenProjects() {
+        return mavenSession.getProjects().stream()
+                .filter(o -> o != mavenSession.getCurrentProject())
+                .map(o -> o.getBasedir().toPath())
+                .collect(Collectors.toSet());
+    }
+
     public List<SourceFile> listSourceFiles(Iterable<NamedStyles> styles,
                                             ExecutionContext ctx) throws DependencyResolutionRequiredException, MojoExecutionException {
 
-        Set<Path> alreadyParsed = new HashSet<>();
         JavaParser javaParser = JavaParser.fromJavaVersion()
                 .styles(styles)
                 .logCompilationWarningsAndErrors(false)
@@ -321,6 +327,7 @@ public class MavenMojoProjectParser {
                 generatedSourcePaths.stream(),
                 listJavaSources(mavenProject.getBuild().getSourceDirectory()).stream()
         ).collect(toList());
+        Set<Path> alreadyParsed = new HashSet<>(mainJavaSources);
 
         List<SourceFile> sourceFiles = new ArrayList<>();
         Xml.Document maven = parseMaven(ctx);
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/maven/ResourceParser.java
```diff
@@ -106,4 +111,55 @@ public <S extends SourceFile> List<S> parseSourceFiles(
             throw new UncheckedIOException(e);
         }
     }
+
+    public <S extends SourceFile> List<S> parseQuarks(
```
tkvangorder, line 115: I believe there is a static method on the QuarkParser that you can use rather than rolling your own. 

The only issue here is that the static method does not have the ability to pass a separate `baseDir` and `searchDir` and it does not allow an ExecutionContext to be passed to it. 

I would just overload the static method in QuarkParser to add these two parameters. 

This method becomes much simplier: 

```
List<Quark> quarks = QuarkParser.parseAllOtherFiles(baseDir, searchDir, alreadyParsed, ctx);
alreadyParsed.addAll(quarks.stream.map(SourceFile::getPath).collect(Collectors.toSet()));
return (List<S>) quarks;
```

tkvangorder (changes requested):
Let me know if you have any questions.

src/main/java/org/openrewrite/maven/ResourceParser.java
```diff
@@ -90,8 +92,7 @@ public <S extends SourceFile> List<S> parseSourceFiles(
 
             long fileSize = attrs.size();
             if ((sizeThresholdMb > 0 && fileSize > sizeThresholdMb * 1024L * 1024L)) {
-                alreadyParsed.add(path);
-                logger.info("Skipping parsing " + path + " as its size + " + fileSize / (1024L * 1024L) +
+                logger.info("Parsing as Quark " + path + " as its size + " + fileSize / (1024L * 1024L) +
                         "Mb exceeds size threshold " + sizeThresholdMb + "Mb");
                 return false;
```
sambsnyd, line 97: You should parse it as a quark right here and add it to the `alreadyParsed` list or this message will be displayed n times, where "n" is the number of parsers that come before the quark parser.

src/main/java/org/openrewrite/maven/ResourceParser.java
```diff
@@ -106,4 +107,56 @@ public <S extends SourceFile> List<S> parseSourceFiles(
...
+    public <S extends SourceFile> List<S> parseQuarks(
+            Path baseDir,
+            Path searchDir,
+            Collection<Path> alreadyParsed,
+            ExecutionContext ctx) {
+        QuarkParser parser = new QuarkParser();
+        try (Stream<Path> resources = Files.find(searchDir, 16, (path, attrs) -> {
+            if (path.toString().endsWith(".java")) {
+                return false;
+            }
```
sambsnyd, line 120: Why is this necessary?

sambsnyd (changes requested):

src/main/java/org/openrewrite/maven/ResourceParser.java
sambsnyd: Alternatively, refactoring `ResourceParser` to not re-traverse the search directory for every single parser could also address this.

src/main/java/org/openrewrite/maven/ResourceParser.java
sambsnyd: Oh I see we run `ResourceParser` before `JavaParser`. Makes sense

traceyyoshima: I ran the maven plugin on a few projects and java files are added without the condition.

- `.java` files aren't added to already parsed. I tested using `listJavaSources` to add the paths as already parsed, but the files were still added as quarks.
- The other resource parsers only accept specific types of files `.yaml`, `.java`, `.xml`, but the `QuarkParser#accept()` method always returns `true`.

tkvangorder: I think a better, long-term solution, would just be to map the paths for all the Java source files into the `alreadyParsed` and not have any specialized logic. I envision a world where there will also be groovy and Kotlin parsers in the mix and I would not want to have separate logic to exclude certain file masks (like "*.java")

tkvangorder: We do parse the Java sources before the resources (I believe)

tkvangorder: Is there something specialized in `parseQuarks` that is not handled by `parseSourceFiles`? It seems largely a duplicate of that method.

src/main/java/org/openrewrite/maven/ResourceParser.java
```diff
@@ -90,17 +87,77 @@ public <S extends SourceFile> List<S> parseSourceFiles(
 
             long fileSize = attrs.size();
             if ((sizeThresholdMb > 0 && fileSize > sizeThresholdMb * 1024L * 1024L)) {
-                alreadyParsed.add(path);
-                logger.info("Skipping parsing " + path + " as its size + " + fileSize / (1024L * 1024L) +
+                logger.info("Parsing as Quark " + path + " as its size + " + fileSize / (1024L * 1024L) +
                         "Mb exceeds size threshold " + sizeThresholdMb + "Mb");
-                return false;
+                return true;
             }
```
sambsnyd, line 93: Size threshold doesn't work anymore. You return `true` here but there's no logic below to make sure that gets parsed as a quark

sambsnyd (changes requested):

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -306,10 +306,16 @@ private static RawRepositories buildRawRepositories(@Nullable List<Repository> r
         return rawRepositories;
     }
 
+    private Set<Path> skipOtherMavenProjects() {
+        return mavenSession.getProjects().stream()
```
traceyyoshima, line 310: Note for posterity:
`mavenSession#getProjects()` returns the projects that are already sorted in reactor order.

```java
    @Override
    public void execute() throws MojoExecutionException {
        if (mavenSession.getCurrentProject == mavenSession.getProjects().get(mavenSession.getProjects().size() - 1)) {
           // do something when the last project is building.
        }
    }
```

src/main/java/org/openrewrite/maven/ResourceParser.java
```diff
@@ -3,107 +3,168 @@
...
+            } else if (yamlParser.accept(path)) {
+                yamlPaths.add(path);
+            } else if (propertiesParser.accept(path)) {
+                propertiesPaths.add(path);
+            } else if (protoParser.accept(path)) {
+                protoPaths.add(path);
+            } else if (hclParser.accept(path)) {
+                hclPaths.add(path);
+            } else if (quarkParser.accept(path)) {
+                quarkPaths.add(path);
```
traceyyoshima, line 130: Java files that do not exist in a maven module or are in a module but exist outside of `main/test` are added as quarks. We could use the JavaParser to parse those as resources if preferred.

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -342,11 +349,11 @@ public List<SourceFile> listSourceFiles(Iterable<NamedStyles> styles,
         sourceFiles.addAll(ListUtils.map(maybeAutodetectStyles(javaParser.parse(mainJavaSources, baseDir, ctx), styles),
                 addProvenance(baseDir, projectProvenance, generatedSourcePaths)));
 
-        ResourceParser rp = new ResourceParser(logger, exclusions, sizeThresholdMb);
+        ResourceParser rp = new ResourceParser(baseDir, logger, exclusions, sizeThresholdMb, skipOtherMavenProjects());
 
         // Any resources parsed from "main/resources" should also have the main source set added to them.
         sourceFiles.addAll(ListUtils.map(
-                rp.parse(baseDir, mavenProject.getBasedir().toPath().resolve("src/main/resources"), alreadyParsed),
+                rp.parse(mavenProject.getBasedir().toPath().resolve("src/main/resources"), alreadyParsed),
```
tkvangorder, line 356: I am a little concerned here because the `baseDir` is the root of the maven project and all paths are made relative to it....this helps when we are checking for exclusions...

To clarify here the instance variable `baseDir` should be the root directory for the entire maven project and `mavenProject.getBaseDir()` will be the directory of the current project.

I could be wrong here and maybe you are accounting for this below....just an observation.

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -306,10 +306,16 @@ private static RawRepositories buildRawRepositories(@Nullable List<Repository> r
         return rawRepositories;
     }
 
+    private Set<Path> skipOtherMavenProjects() {
+        return mavenSession.getProjects().stream()
+                .filter(o -> o != mavenSession.getCurrentProject())
+                .map(o -> o.getBasedir().toPath())
```
tkvangorder, line 312: Same thing here, shouldn't these paths be resolved from `baseDir` (the instance variable)?

traceyyoshima:
Closing the PR to separate changes from adding Quark support and #352.

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
traceyyoshima: The `baseDir` is passed into the ResourceParser constructor on line 352.

Prior to the change, PathMatchers were generated using `baseDir.getFileSystem().getPathMatcher(..)` per path that is traversed. By moving the `baseDir` into the constructor, we don't need to pass the implied path into each method call. And the PathMatchers may be generated during instantiation of ResourceParser rather than doing the work repeatedly.

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
traceyyoshima: No, I don't think so.

PR #354, opened 2022-05-06
Collect maven poms with LinkedHashSet.

### LINKED ISSUE #351: rewriteRun adds a `Maven` Xml.Document with a different sourcePath than the current MavenProject
Maven builds multi-module projects in project reactor order.
As each project is built the `mavenProject` is equivalent to `mavenSession.getCurrentProject`.

The `Xml.Document` returned from the `MavenParserBuilder` typically does not match the current maven project pom. AFAICT the XML document that is returned always has the same source path.

```yaml
root:
  module1:
    - pom1.xml # hypothetical example of what is returned from MavenParserBuilder.
  module2:
    - pom2.xml # parser builder will return pom1.xml, so the resource parse will pick up pom2.xml without markers.
```

traceyyoshima:
Changes:

- parseMaven will return a `Maven` `Xml.Document` that matches the current project.

fixes #351

### DIFF
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -145,7 +145,7 @@ public class MavenMojoProjectParser {
             return null;
         }
 
-        Set<Path> allPoms = collectPoms(mavenProject, new HashSet<>());
+        Set<Path> allPoms = collectPoms(mavenProject, new LinkedHashSet<>());
         mavenSession.getProjectDependencyGraph().getUpstreamProjects(mavenProject, true).forEach(p -> collectPoms(p, allPoms));
         MavenParser.Builder mavenParserBuilder = MavenParser.builder().mavenConfig(baseDir.resolve(".mvn/maven.config"));
 


src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -145,7 +145,7 @@ public Xml.Document parseMaven(ExecutionContext ctx) {
             return null;
         }
 
-        Set<Path> allPoms = collectPoms(mavenProject, new HashSet<>());
+        Set<Path> allPoms = collectPoms(mavenProject, new LinkedHashSet<>());
```
traceyyoshima, line 148: Note:
The MavenParser maintains the order of poms with a LinkedHashMap, but we are passing in a `HashSet`. The HashSet causes the order to be based on the hashcode so `.iterator().next()` typically returns the wrong value. Then the resource parser adds the pom as a normal `Xml.Document`, because the `maven.getSourcePath` does not match the current project.

Alternatively, we can use a `HashSet` here, and refactor `parseMaven` to find the `Xml.Document` from the list of results returned by the MavenParser.

```java
        Xml.Document maven = mavenParserBuilder
                .build()
                .parse(allPoms, baseDir, ctx)
                .iterator() // Could stream the results and find a match based on the current project
                .next();
```

But I opted for a `LinkedHashSet` for simplicity and because there may be a lot of `Xml.Documents` to scan through, so this seems more efficient.

PR #359, opened 2022-05-10
Refactoring to the plugin to only parse and run on the last project. fixes #349

### LINKED ISSUE #349: Maven plugin should apply recipe to all sub modules in one pass
Currently each sub-module is parsed independently of each other, and recipe runs are applied only to a sub-module. 
It is currently impossible for a `Recipe` run by the maven plugin to look at information in sub-module A and use that information while visiting a class in sub-module B.

This behavior differs from our Gradle plugin. The Gradle plugin is able to aggregate all subprojects together into a single list of ASTs and `Recipe`s are applied to them together, rather than independently.

I'm not sure how to configure Maven to do this.

tkvangorder:
Reworked the maven plugin to run on the last project in the maven session. The plugin will then iterate through all of the projects in the maven session (in reactor order) to parse source files and will then run the recipe across all source files (in all projects).

### DIFF
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -18,6 +18,7 @@ import org.openrewrite.SourceFile;
 import org.openrewrite.internal.ListUtils;
 import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.java.JavaParser;
+import org.openrewrite.java.internal.JavaTypeCache;
 import org.openrewrite.java.marker.JavaProject;
 import org.openrewrite.java.marker.JavaVersion;
 import org.openrewrite.java.style.Autodetect;
@@ -70,6 +71,8 @@ public class MavenMojoProjectParser {
 
     @Nullable
     static MavenPomCache pomCache;
+
+    static JavaTypeCache typeCache = new JavaTypeCache();
 
     private final Log logger;
     private final Path baseDir;
@@ -132,6 +135,7 @@ public class MavenMojoProjectParser {
 
         JavaParser javaParser = JavaParser.fromJavaVersion()
                 .styles(styles)
+                .typeCache(typeCache)
                 .logCompilationWarningsAndErrors(false)
                 .build();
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -124,21 +129,95 @@ public MavenMojoProjectParser(Log logger, Path baseDir, boolean pomCacheEnabled,
...
-        } catch (Exception e) {
-            // Logging at a low level as this is unlikely to happen except in non-git projects, where it is expected
-            logger.debug("Unable to determine git provenance", e);
+        JavaParser javaParser = JavaParser.fromJavaVersion()
+                .styles(styles)
+                .logCompilationWarningsAndErrors(false)
+                .build();
+
+        // Some annotation processors output generated sources to the /target directory
+        List<Path> generatedSourcePaths = listJavaSources(mavenProject.getBuild().getDirectory());
```
jkschneider, line 139: If we parse generated sources as ASTs, we run the risk of generating diffs that don't accomplish anything when making a git commit.

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -73,29 +73,34 @@ public class MavenMojoProjectParser {
...
         this.pomCacheDirectory = pomCacheDirectory;
         this.skipMavenParsing = skipMavenParsing;
+        this.buildTool = new BuildTool(randomId(), BuildTool.Type.Maven, runtime.getMavenVersion());
         this.exclusions = exclusions;
         sizeThresholdMb = thresholdMb;
         this.mavenSession = session;
         this.settingsDecrypter = settingsDecrypter;
+    }
+
+    public List<SourceFile> listSourceFiles(MavenProject mavenProject, Iterable<NamedStyles> styles,
```
jkschneider, line 102: this method is starting to feel uncomfortably long, like one of those things that you are afraid to touch

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
tkvangorder: oh, great catch.

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
tkvangorder: Yeah, I am going to clean this up and extract logic into easier-to-reason-about methods.

tkvangorder:
I am going to merge this because Tracey and I are iterating on this together. We will address your feedback @jkschneider

PR #384, opened 2022-06-20
Upping maven-filter from 3.2.0 (Aug 2020) to 3.3.0 (Jun, 2022)

rassmate:
Fixing a null value in the activeRecipes collection causing the discover goal to fail even if no recipes should be run. This can happen when having a property as the value of the recipe name and that property is null. Then the collection will only contain a single element which is null

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -170,8 +170,9 @@
         <dependency>
             <groupId>org.apache.maven.shared</groupId>
             <artifactId>maven-filtering</artifactId>
-            <version>3.2.0</version>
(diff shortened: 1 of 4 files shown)


pway99:
Thanks @rassmate !

rassmate:
Thanks for the merge.
Any chance of getting this in a release anytime soon @pway99?

pway99:
Hi @rassmate, It's hard to say; we try and release every few weeks, 14 days since the last release with 36 commits since then.  So I would expect a release in the next week or two.

Issue #386, opened 2022-06-23
Discover goal behaviour is not optimal for multi-module projects / no projects


dsibilio:
When trying to run `mvn rewrite:discover` in a multi-module project, the goal will be repeated for each module. This seems to be unnecessary since the same output gets produced over and over, and can actually be quite annoying - especially if the goal has been run in interactive mode.

When trying to run `mvn org.openrewrite.maven:rewrite-maven-plugin:4.26.0-SNAPSHOT:discover` in a folder that's not a Maven project, the execution will fail - even though the target project doesn't seem to be important for the execution of the goal.

For comparison purposes, `mvn help:describe` (maven-help-plugin) does both things the opposite way, executing the goal only once even for multi-module projects, and allowing execution from within non-Maven project folders.

PR #387, opened 2022-06-23
Improve 'discover' goal behaviour for multi-module/no projects

dsibilio:
Allows for the execution of the `discover` goal from non-Maven projects.
Will only execute the `discover` goal once per project reactor rather than once for each module.

Fixes https://github.com/openrewrite/rewrite-maven-plugin/issues/386.


sambsnyd:
Thanks @dsibilio !

Issue #437, opened 2022-11-14
-Drewrite.recipeArtifactCoordinates should include transitive dependencies


ge0ffrey:
In the maven plugin, the command line option `rewrite.recipeArtifactCoordinates` does not take into account transitive dependencies, causing recipes (that are not part of rewrite upstream, so not AutoFormat)  to thrown an error.

**Reproducer**

1. Run optaplanner-migration on an optaplanner quickstart

```
git clone https://github.com/kiegroup/optaplanner-quickstarts
cd optaplanner-quickstarts
git checkout 8.27.0.Final
cd use-cases/school-timetabling
mvn clean org.openrewrite.maven:rewrite-maven-plugin:4.36.0:run -Drewrite.recipeArtifactCoordinates=org.optaplanner:optaplanner-migration:8.31.0.Final -Drewrite.activeRecipes=org.optaplanner.migration.ToLatest
```
Note: if optaplanner 8.31.0.Final is not out yet, [build optaplanner from scatch](https://github.com/kiegroup/optaplanner) and use 8.31.0-SNAPSHOT instead.

Then open `TimeTableConstraintProvider.java` and check if the roomConflict() method has changed and how.

Here's how it looks like (bad):

![image](https://user-images.githubusercontent.com/176880/201675806-6d026e5a-a242-4abc-9c81-ece5af178abd.png)

This is what it should be (good):

```
    Constraint roomConflict(ConstraintFactory constraintFactory) {
                ...
                // ... and penalize each pair with a hard weight.
                .penalize(HardSoftScore.ONE_HARD);
                .asConstraint("Room conflict");
    }
```

tkvangorder:
Transferred. Thank you for reporting this issue!

ge0ffrey:
Slack info:

@tkvangorder wrote:

>  It appears that we are using a custom class loader (to pick up these artifacts) when loading the environment: https://github.com/openrewrite/rewrite-maven-plugin/blob/main/src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java#L73 So, that would explain why it is finding the recipe. However, I think because we are also importing rewrite classes into the Mojo, these classes are going to be loaded with the plugin's class loader. I honestly didn't know we had added this ability. I think many folks have been using the rewrite:configure to add the plugin to pom.xml and then rewrite:remove to remove it. I know this is not quite the same experience.I think we will have to investigate/unify how class loading is done before this will work correctly.


@kmclarnon wrote:
[...]

PR #445, opened 2022-12-08
Bugfix: transitive dependencies of recipeArtifactCoordinates jars will now be available during recipe execution

nmck257:
#437

### DIFF
--- a/src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
+++ b/src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
@@ -13,13 +13,13 @@ import org.apache.maven.project.MavenProject;
 import org.apache.maven.repository.RepositorySystem;
 import org.apache.maven.rtinfo.RuntimeInformation;
 import org.apache.maven.settings.crypto.SettingsDecrypter;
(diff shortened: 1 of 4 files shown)


src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
```diff
@@ -65,12 +65,16 @@ public abstract class AbstractRewriteMojo extends ConfigurableRewriteMojo {
...
+    protected Environment environment(@Nullable ClassLoader recipeClassLoader) throws MojoExecutionException {
         Environment.Builder env = Environment.builder(project.getProperties());
-        if (getRecipeArtifactCoordinates().isEmpty()) {
+        if (recipeClassLoader == null) {
             env.scanRuntimeClasspath()
                     .scanUserHome();
         } else {
-            env.load(new ClasspathScanningLoader(project.getProperties(), getRecipeClassloader()));
+            env.load(new ClasspathScanningLoader(project.getProperties(), recipeClassLoader));
         }
```
nmck257, line 78: With these changes, we should get the same result just running `env.scanRuntimeClasspath` unconditionally here, but I kept this `if/else` for now to ensure the existing behavior is preserved if we run into the ClassCastException risked in the `merge` method

PR #469, opened 2023-01-15
build: update maven wrapper to 3.8.7

yeikel:
I generated this pull request running `mvn wrapper:wrapper -Dmaven=3.8.7`


timtebeek:
Thanks again @yeikel ! We should really [get that recipe working](https://github.com/openrewrite/rewrite/issues/1565) and then run that periodically. Until then thanks for helping out!

Issue #483, opened 2023-01-23
Stream closed error when using a URL in the config location


gastaldi:
I was testing the introduced support for URLs in the config location and am getting the following error. This is always reproducible. Also, I am not running behind a proxy:

```
Caused by: java.io.IOException: stream is closed
    at sun.net.www.protocol.http.HttpURLConnection$HttpInputStream.ensureOpen (HttpURLConnection.java:3651)
    at sun.net.www.protocol.http.HttpURLConnection$HttpInputStream.read (HttpURLConnection.java:3656)
    at org.openrewrite.ipc.http.HttpSender$Response$1.read (HttpSender.java:332)
    at java.io.InputStream.read (InputStream.java:284)
    at org.openrewrite.config.YamlResourceLoader.<init> (YamlResourceLoader.java:154)
    at org.openrewrite.config.YamlResourceLoader.<init> (YamlResourceLoader.java:118)
    at org.openrewrite.config.YamlResourceLoader.<init> (YamlResourceLoader.java:105)
    at org.openrewrite.maven.AbstractRewriteMojo.environment (AbstractRewriteMojo.java:119)
    at org.openrewrite.maven.AbstractRewriteMojo.listResults (AbstractRewriteMojo.java:185)
    at org.openrewrite.maven.AbstractRewriteRunMojo.execute (AbstractRewriteRunMojo.java:48)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
```

This is the command I executed to reproduce it: 

```bash
mvn org.openrewrite.maven:rewrite-maven-plugin:4.39.0:run \
   -Drewrite.configLocation=https://raw.githubusercontent.com/quarkusio/quarkus/main/jakarta/quarkus3.yml \
   -DactiveRecipes=io.quarkus.openrewrite.Quarkus3
``` 

Related issues: 

- #432

kunli2:
The next version after 4.39.0 will include the fix.

PR #484, opened 2023-01-25
fix stream closed error when using a URL in the config

kunli2:
Fix issue https://github.com/openrewrite/rewrite-maven-plugin/issues/483.
### root cause
The "try-with-resources" block closed the response automatically, so at the consuming place of the `config.inputStream`, when reading from the stream, it has been closed.

### DIFF
--- a/src/test/java/org/openrewrite/maven/AbstractRewriteMojoTest.java
+++ b/src/test/java/org/openrewrite/maven/AbstractRewriteMojoTest.java
@@ -18,7 +18,11 @@ import static org.assertj.core.api.Assertions.assertThat;
 public class AbstractRewriteMojoTest {
 
     @ParameterizedTest
-    @ValueSource(strings = {"rewrite.yml", "https://httpstat.us/200"})
+    @ValueSource(strings = {
+        "rewrite.yml",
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/maven/AbstractRewriteMojoTest.java
```diff
@@ -18,7 +18,11 @@
 public class AbstractRewriteMojoTest {
 
     @ParameterizedTest
-    @ValueSource(strings = {"rewrite.yml", "https://httpstat.us/200"})
+    @ValueSource(strings = {
+        "rewrite.yml",
+        "https://httpstat.us/200",
+        "https://raw.githubusercontent.com/quarkusio/quarkus/main/jakarta/quarkus3.yml"
```
sambsnyd, line 24: Let's host the yaml file in the test case somewhere we control, rather than link to a quarkus github repo which might move at any time. Can check it in to rewrite-maven-plugin directly

kunli2: Sure, let me remove the 3rd test case since the "https://httpstat.us/200" covered this as well.

PR #489, opened 2023-02-06
#412 - Added OS Discovery to Provenance

reka18:


pom.xml
```diff
@@ -64,7 +64,7 @@
     </scm>
 
     <properties>
-        <rewrite.version>7.35.0</rewrite.version>
+        <rewrite.version>7.36.0</rewrite.version>
```
timtebeek, line 67: Were you looking to keep this open until after the 7.36.0 release? Otherwise we can test it now with
```suggestion
        <rewrite.version>7.36.0-SNAPSHOT</rewrite.version>
```

Issue #490, opened 2023-02-06
Link to blog post is wrong


khmarbaise:
The link to the blog post https://blog.soebes.de/blog/2020/08/18/itf-part-i/ is wrong.

timtebeek:
Fixed the link just now; good seeing you here!

Missed you in the hallway track at FOSDEM, but if there's ever anything we can help with let us know :)

khmarbaise:
I've provided a pr to fix this. I had been in the dev room for the openrewrite lecture 👍

timtebeek:
Yes glad to hear! If you ever want to apply some recipes in bulk we have 90 Apache Maven repositories in https://public.moderne.io ; I had done some earlier pull requests manually to apache/maven-wrapper for their JUnit 5 migration; we can scale that up if you'd like. I'm available in the Foojay Slack if it helps.

PR #491, opened 2023-02-06
fix link to blog post ist wrong. (#490)

khmarbaise:


timtebeek:
Locally I only had success opening https://blog.soebes.io/posts/2020/08/2020-08-17-itf-part-i/
Would you agree that link works, whereas https://blog.soebes.de/posts/2020/08/18/itf-part-i/ gives a 404?

Issue #496, opened 2023-02-06
Remove public modifier from integration tests classes


khmarbaise:
Currently all integration tests are defined like:
```java
..
public class BasicIT {
..
}
```
which is not necessary.

timtebeek:
That's embarrassing; [we even have a recipe for that](https://docs.openrewrite.org/reference/recipes/java/testing/cleanup/testsshouldnotbepublic); let me see if I can run that quickly. :)

timtebeek:
Thanks again! We have a plan internally to apply our recipes to our own code continuously; thanks for the reminder to move that up in priority!

PR #497, opened 2023-02-06
fix Remove public modifier from integration tests classes (#496)

khmarbaise:
* remove also unused class TimeUtils.


timtebeek:
Thanks again!

PR #505, opened 2023-02-11
fix - BOM Usage instead of separate dependencies with properties. (#504)

khmarbaise:
- remove junit-jupiter.version property because we have only a single location.


timtebeek:
Much appreciated yet again!

PR #515, opened 2023-02-23
feat: Add rewrite.skip to skip execution

berlam:
Just a simple way to skip plugin execution by property to allow more granular control in the lifecycle without profiles.


jkschneider:
Thanks, makes sense!

sohailalam2:
Thank you :) @jkschneider 
Looks like this is not included in v4.40.0. When is this fix scheduled for release?

PR #517, opened 2023-03-02
Do not write to quark

timtebeek:
Fixes https://github.com/openrewrite/rewrite-cloud-suitability-analyzer/issues/6


timtebeek:
Also reported in https://stackoverflow.com/questions/75630973/openrewrite-the-contents-of-a-quark-are-unknown-so-the-charset-is-unknown

PR #549, opened 2023-05-05
OpenRewrite 8.0 upgrade

knutwannheden:


timtebeek:
On 7.x developers can run a lot of static analysis recipes without adding any dependencies, with a single Maven command. For instance:

    mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.activeRecipes=org.openrewrite.java.cleanup.EqualsAvoidsNull

In 8.x a lot of these clean up recipes have been moved to rewrite-static-analysis, a separate module that requires an explicit dependency. Would we want to include that dependency by default here to keep that option?

knutwannheden:
> Would we want to include that dependency by default here to keep that option?

Could make sense. But then we would probably have to create a new release of the Maven plugin every time we release `rewrite-static-analysis` as well, as otherwise the instruction from the docs won't work. Right?

timtebeek:
Guess you're right; I think if we don't do anything the docs will start to generate new code snippets that include the dependency, so we can start with that and see if we get any feedback. Or have the plugin start to use the recipe-bom, with optional cascading releases on that. 🤔

PR #552, opened 2023-05-08
Use a defined order for projectMap

timtebeek:
LinkedHashMap instead of HashMap, to mimic https://github.com/openrewrite/rewrite/commit/8d7eff280c016aa7c47238898062c29457f81ce4

https://github.com/openrewrite/rewrite/issues/1753
https://github.com/openrewrite/rewrite/issues/3219

### DIFF
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -357,7 +357,7 @@ public class MavenMojoProjectParser {
         MavenProject topLevelProject = mavenSession.getTopLevelProject();
         logInfo(topLevelProject, "Resolving Poms...");
 
-        Set<Path> allPoms = new HashSet<>();
+        Set<Path> allPoms = new LinkedHashSet<>();
         mavenProjects.forEach(p -> collectPoms(p, allPoms));
         for (MavenProject mavenProject : mavenProjects) {
             mavenSession.getProjectDependencyGraph().getUpstreamProjects(mavenProject, true).forEach(p -> collectPoms(p, allPoms));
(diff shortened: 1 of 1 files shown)


timtebeek:
Going on an instinct here, and this has not been tested. Similar change based off on what I saw in https://github.com/openrewrite/rewrite/pull/3155

rpau (approved):

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -402,7 +402,7 @@ public Map<MavenProject, Xml.Document> parseMaven(List<MavenProject> mavenProjec
         }
 
         Map<Path, MavenProject> projectsByPath = mavenProjects.stream().collect(Collectors.toMap(MavenMojoProjectParser::pomPath, Function.identity()));
-        Map<MavenProject, Xml.Document> projectMap = new HashMap<>();
+        Map<MavenProject, Xml.Document> projectMap = new LinkedHashMap<>();
```
timtebeek, line 405: Might need to move this up still; order is defined further up.

timtebeek:
This is based on the knowledge that the input projects from `mavenSession.getProjects()` [here](https://github.com/openrewrite/rewrite-maven-plugin/blob/379968c7788617dfe89d2a9531e722853c15d53c/src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java#LL276C100-L276C111) has been sorted:

    /**
     * These projects have already been topologically sorted in the {@link org.apache.maven.Maven} component before
     * being passed into the session. This is also the potentially constrained set of projects by using --projects
     * on the command line.
     */
    private List<MavenProject> projects;

PR #554, opened 2023-05-10
Fixed NPE

nmck257:
For obscure case where maven-compiler-plugin's source/target/release config is declared without a value


timtebeek (approved):

timtebeek (approved):

timtebeek:
Thanks again!

PR #556, opened 2023-05-14
Bump rewrite.version properties on release

timtebeek:
This ensures we use the latest versions of rewrite modules with each release, in lieu of using `RELEASE` which made it into the released pom.xml.


knutwannheden:
Sounds good. Is this a separate workflow? Maybe we could also have one to upgrade to the latest snapshot after making a new release.

timtebeek:
This worked: https://github.com/openrewrite/rewrite-maven-plugin/commit/0667ddd0b4910f579cbeeebfada9bbf55ff39a36

PR #569, opened 2023-06-02
support rewrite.inactiveRecipes

JoeCqupt:
## What's changed?

support rewrite.inactiveRecipes

## What's your motivation?

when i use  a recipe (it consist of many recipe), maybe i want disable some leaf-recipe.

please focus on is this idea is right? 

anyone

> Recipe class add property `enable`. when `enable` is false this Recipe not work. 
>  
> `rewrite-maven-plugin` set param `rewrite.inactiveRecipes ` .
> 
> `rewrite-maven-plugin` find those recipes to recipe.setEnable(false)

is this idea better? but implement code difficult

## Any additional context
none
### Checklist
- [x] I've added the license header to any new files through `./gradlew licenseFormat`
- [x] I've used the IntelliJ auto-formatter on affected files
- [x] I've updated the documentation (if applicable)

### DIFF
--- a/src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
+++ b/src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
@@ -239,6 +239,7 @@ public abstract class AbstractRewriteMojo extends ConfigurableRewriteMojo {
             }
 
             Recipe recipe = env.activateRecipes(getActiveRecipes());
+            inactiveRecipes(recipe);
             if (recipe.getRecipeList().isEmpty()) {
                 getLog().warn("No recipes were activated. " +
                               "Activate a recipe with <activeRecipes><recipe>com.fully.qualified.RecipeClassName</recipe></activeRecipes> in this plugin's <configuration> in your pom.xml, " +
@@ -297,6 +298,37 @@ public abstract class AbstractRewriteMojo extends ConfigurableRewriteMojo {
             return new ResultsContainer(repositoryRoot, results);
         } catch (DependencyResolutionRequiredException e) {
             throw new MojoExecutionException("Dependency resolution required", e);
+        }
+    }
+
+    private void inactiveRecipes(Recipe rootRecipe) {
+        // processed recipe
+        Set<Recipe> processed = new HashSet<>();
+        Queue<Recipe> queue = new LinkedList<>();
+        queue.offer(rootRecipe);
+        while (!queue.isEmpty()) {
+            Recipe recipe = queue.poll();
+
+            if (processed.contains(recipe)) {
+                continue;
+            }
+
+            Iterator<Recipe> iterator = recipe.getRecipeList().iterator();
+            while (iterator.hasNext()) {
+                Recipe r = iterator.next();
+                if (getInactiveRecipes().contains(r.getName())) {
+                    try {
+                        iterator.remove();
+                    } catch (UnsupportedOperationException uoe) {
+                        // maybe unmodifiableCollection
+                        getLog().warn("InactiveRecipe " + r.getName() + " for " +
+                                recipe.getName() + " failed. ", uoe);
+                    }
+                }
+            }
+
+            recipe.getRecipeList().stream().forEach(r -> queue.offer(r));
+            processed.add(recipe);
         }
     }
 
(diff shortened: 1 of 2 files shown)


timtebeek:
Hi @JoeCqupt ! Thank you for looking into this; it's indeed a frequent request to be able to exclude one or more recipes, which we have so far pushed back upon. The reason is it's typically possible to work around such issues by [creating a separate `rewrite.yml` description file](https://docs.openrewrite.org/reference/yaml-format-reference), which only contains the desired recipes.

It's more complicated than the current implementation to reliably implements exclusions, as recipes can also directly invoke each other, which is becoming quite common as well. I believe such direct invocations would not be picked up with the current implementation, and would not be trivial to implement.

There's also potential for issues when recipes assume preceding recipes have run before they make their own changes, which will be hard to reliably detect and prevent. In the past for instance we've had to request not to migrate just yet [to JUnit Jupiter as part of the Spring Boot 2.4 migration](https://github.com/openrewrite/rewrite-spring/blob/b5fb4ea471c972474699bde6668e2fac5e065871/src/main/resources/META-INF/rewrite/spring-boot-24.yml#L139-L140). If we would allow exclusions, then we suddenly can't be certain anymore that users are on JUnit Jupiter in any of the Spring Boot 2.4+ recipes, which really complicates the logic for all recipes involved. There's more such examples where exclusions might trip up migrations in unexpected ways, which is why we've found it easier not to support migrations rather than open it up for potential misuse, which we then have to counteract.

We do try to be conservative with the recipes we include with our common larger composite recipes, such that there should rarely be a need for exclusions, especially with the `rewrite.yml` workaround available as well. When issues do occur with included recipes, we prefer to fix those bugs quickly rather than allow for exclusions that might harder to debug issues down the line.

So as much as we appreciate you contributing to the project, at this stage I'm uncertain whether this particular propose is one we would be open to adopt. I'll raise it again internally, but wanted to give you this background such that you know it's both more complicated than expected, and has a potential for further issues that we hope to prevent.

Feel free to add your own perspective here as well; you likely have a couple of use cases that you might have hoped to use this feature for, and I hope we are able to work through those for you in some other way!

JoeCqupt:
ok. i get it.   i can resolve problem by creating `rewrite.yml`.

timtebeek:
Yes sorry to see you've spent time on something we won't continue with for now. You're always welcome to contribute in some other way! You can also [join our slack](https://join.slack.com/t/rewriteoss/shared_invite/zt-nj42n3ea-b~62rIHzb3Vo0E1APKCXEA) if there's anything you'd want to discuss up front before diving in. And like I said before: happy to work through any issues that prompted you to consider exclusions!

tgeens:
Maybe there is a possibility to provide an optional `.enabled` configuration flag for recipes that are rather optional ?

For example: `UpgradeSpringBoot_3_1` transitively pulls in `org.openrewrite.java.spring.ImplicitWebAnnotationNames`, removing explicitly named request-parameters from RestController signatures.  If this goes against my projects' style guide, I might be better off _not_ using OpenRewrite ?

timtebeek:
Hi @tgeens ; Sorry to hear you're unhappy with the inclusion of [`ImplicitWebAnnotationNames`](https://docs.openrewrite.org/recipes/java/spring/implicitwebannotationnames). For reference that comes in through [our Spring Boot 2 best practices](https://github.com/openrewrite/rewrite-spring/blob/56d60e00da710df3b5132fc17fa319e85cd20e8a/src/main/resources/META-INF/rewrite/spring-boot-20.yml#L671-L679), which we run as [part of our Migration to Spring Boot 2.0](https://github.com/openrewrite/rewrite-spring/blob/56d60e00da710df3b5132fc17fa319e85cd20e8a/src/main/resources/META-INF/rewrite/spring-boot-20.yml#L68). That's indeed potentially a lot of yaml recipes to copy if you were to follow the above recommended approach.

It's not immediately clear to me which recipes you'd like to add a `.enabled` configuration option to; it sounds very much like a skip option by another name to me, which would suffer from the same drawbacks as outlined above.

Perhaps you'd be interested in following this PR, which aims to allow you to [run individual minor Spring release migrations in isolation](https://github.com/openrewrite/rewrite-spring/pull/361). That could provide you with an alternative here, as you likely already made the leap from 1.5 to 2.x+. Could that be a way forward for you?

timtebeek:
These days you can create your own custom recipe based off of an existing recipe most easily through [the Moderne recipe builder](https://app.moderne.io/builder). You can [read more about the new recipe builder](https://www.moderne.ai/blog/moderne-engineering-building-the-openrewrite-recipe-builder) on the Moderne blog. We think this is the best way forward to allow for customization. Hope that helps!

Issue #566, opened 2023-05-31
Wrong version of `io.netty:netty-handler` is reported in the SBOM of `Selenese Runner Java`


algomaster99:
## What version of OpenRewrite are you using?

I am using

- Maven/Gradle plugin v4.45.0

## How are you running OpenRewrite?

I am running the  maven plugin on https://github.com/vmi/selenese-runner-java/tree/3e84e8e4e7e06aa1bdacaa8266db00f62ebef559.
```sh
mvn org.openrewrite.maven:rewrite-maven-plugin:4.45.0:cyclonedx --fail-at-end
```

## What is the smallest, simplest way to reproduce the problem?

```sh
git clone git@github.com:vmi/selenese-runner-java.git
git checkout 3e84e8e4e7e06aa1bdacaa8266db00f62ebef559
mvn org.openrewrite.maven:rewrite-maven-plugin:4.45.0:cyclonedx --fail-at-end
# convert to json cyclonedx file (if needed)
# I needed JSON format so I used https://github.com/CycloneDX/cyclonedx-cli/releases/download/v0.24.2/cyclonedx-linux-x64
```

## What did you expect to see?

The version of `io.netty:netty-handler` should be `4.1.79.Final` according to [maven dependency plugin](https://maven.apache.org/plugins/maven-dependency-plugin/tree-mojo.html). See our report generated from maven-dependency-plugin [here](https://github.com/chains-project/SBOM-2023/blob/major-revision/sbom-production/results/selenese-runner-java/maven-dependency-tree/tree.json).

## What did you see instead?

`4.1.78.Final` was reported.

## What is the full stack trace of any errors you encountered?

See the SBOM file produced.

https://github.com/chains-project/SBOM-2023/blob/major-revision/sbom-production/results/selenese-runner-java/openrewrite/sbom.json

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?

I could give it a try if the contributors can help me debug it.

joanvr:
Thanks for reporting this issue! Sounds like a big problem if we are not reporting the right version!

Maybe a good way to start working on this would be to try to replicate this in a test like those ones: https://github.com/openrewrite/rewrite-maven-plugin/blob/main/src/test/java/org/openrewrite/maven/RewriteCycloneDxIT.java

algomaster99:
I am on it. I will submit a pull request soon.

[...]

PR #570, opened 2023-06-02
Add test to reproduce wrong version of netty-handler

algomaster99:
## What's changed?

- Added a test
- Added a submodule https://github.com/vmi/selenese-runner-java/tree/3e84e8e4e7e06aa1bdacaa8266db00f62ebef559

## What's your motivation?

Reproduce #566 .

The version of `io.netty:netty-handler` reported is `4.1.78.Final`, which is incorrect based on `maven-dependency-plugin:3.4.0:tree` as reported [here](https://github.com/chains-project/SBOM-2023/blob/major-revision/sbom-production/results/selenese-runner-java/maven-dependency-tree/tree.json?rgh-link-date=2023-05-31T17%3A48%3A09Z). However, I looked up where this dependency comes from, and it is a direct dependency of https://mvnrepository.com/artifact/io.netty/netty-codec-http/4.1.78.Final. Surprisingly, maven central also reports the version as `4.1.78.Final`.

@joanvr 
## Any additional context

I am actually doing a study of SBOM producers and we included your tool in our analysis. You can find the full technical report here - https://arxiv.org/pdf/2303.11102.pdf. We came across this variance in behaviour while looking through different SBOMs.

### Checklist
- [x] I've used the IntelliJ auto-formatter on affected files

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          submodules: true
       - uses: actions/setup-java@v3
         with:
           distribution: temurin
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "src/test/resources-its/org/openrewrite/maven/RewriteCycloneDxIT/correct_version_of_netty_handler_should_be_reported_in_selenese_runner_java_3e84e8e"]
+	path = src/test/resources-its/org/openrewrite/maven/RewriteCycloneDxIT/correct_version_of_netty_handler_should_be_reported_in_selenese_runner_java_3e84e8e
+	url = git@github.com:vmi/selenese-runner-java.git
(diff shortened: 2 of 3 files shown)


timtebeek:
Hi @algomaster99 ! Appreciate you took the time to inform us in detail of this issue with our SBOM report. For inclusion in our project I hope there's a way to replicate this without the need for a git submodule, although I'd understand it if you're not inclined to reduce this further on your own time. Know that we're thankful of the report and hope to have this resolved.

algomaster99:
Hi @timtebeek !

Sorry for the late response, but it has been holiday season here in Sweden. However, I can give a try to generate a minimal example, but I am not sure why this is happening firstly. My only guess right now is that the dependency tree contains two different versions of `netty-handler`, and maven just chooses the latest one.If this is the case, an SBOM producer should report both versions.

timtebeek:
No problem at all & thank you for getting back to us whenever it fits your schedule. It's an interesting theory, and even if there are multiple versions we should still match what Maven does for our reports. Hope we can reproduce this in a smaller sample which we can then ideally fit into our unit tests.

timtebeek:
Closing this one as discussed: ideally we replicate this without the need for a submodule. The Mojo here uses a utility in openrewrite/rewrite, so any improvements are likely to go there as well:
https://github.com/openrewrite/rewrite/blob/v8.10.1/rewrite-maven/src/main/java/org/openrewrite/maven/utilities/PrintMavenAsCycloneDxBom.java

Thanks again for bringing forth this issue!

PR #576, opened 2023-06-06
Implement streaming for source file parsing

knutwannheden:
While the recipe running will still load all parsed source files into memory (by collecting the stream into a List), the parser API can be clients which require streaming support.

### DIFF
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -18,7 +18,6 @@ import org.codehaus.plexus.util.xml.Xpp3Dom;
 import org.jetbrains.annotations.NotNull;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.SourceFile;
-import org.openrewrite.internal.ListUtils;
 import org.openrewrite.internal.StringUtils;
 import org.openrewrite.internal.lang.Nullable;
 import org.openrewrite.java.JavaParser;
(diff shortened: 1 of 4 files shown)


src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -505,7 +502,7 @@ public MavenSettings buildSettings() {
                                 ),
                                 buildRawRepositories(p.getRepositories())
                         )
-                ).collect(toList()));
+                ).collect(Collectors.toList()));
```
timtebeek, line 505:
```suggestion
                ).collect(toList()));
```

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -360,14 +357,14 @@ public Map<MavenProject, Xml.Document> parseMaven(List<MavenProject> mavenProjec
...
-        List<String> activeProfiles = topLevelProject.getActiveProfiles().stream().map(Profile::getId).collect(Collectors.toList());
+        List<String> activeProfiles = topLevelProject.getActiveProfiles().stream().map(Profile::getId).collect(toList());
         if (!activeProfiles.isEmpty()) {
             mavenParserBuilder.activeProfiles(activeProfiles.toArray(new String[]{}));
         }
 
         List<Xml.Document> mavens = mavenParserBuilder
                 .build()
-                .parse(allPoms, baseDir, ctx).collect(toList());
+                .parse(allPoms, baseDir, ctx).collect(Collectors.toList());
```
timtebeek, line 367: Here you're slightly inconsistent as compared to elsewhere in the same file; figured use the same everywhere.
```suggestion
                .parse(allPoms, baseDir, ctx).collect(toList());
```

timtebeek (approved):

PR #577, opened 2023-06-06
Remove unwanted characters from error message

Bananeweizen:
## What's changed?
The error message for resources with parse errors.

## What's your motivation?
The error message currently says "There were problems parsing 7 + sources:".

Looking at the git history the plus sign in the output is probably from a string concatenation that made it into the string literal by accident.

### Checklist
n.a.


timtebeek:
I'm going to agree with you here that the message is sub optimal. Ideally I think we additionally run [Slf4jBestPractices](https://public.moderne.io/recipes/org.openrewrite.java.logging.slf4j.Slf4jBestPractices?) to parameterize the logging statement instead of using String concatenation; should be an easy follow up after this is merged. Thanks for pointing it out!

Issue #593, opened 2023-07-04
`java.specification.version` should be used when no java version for Maven source and target can be retrieved


fabapp2:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.1.2
- Maven plugin v5.2.2

## How are you running OpenRewrite?

Programmatically calling `MavenMojoProjectParser.generateProvenance(MavenProject)` 

## What is the smallest, simplest way to reproduce the problem?
1. Instantiate a `MavenMojoProjectParser` with null and default values
2. Create a `new MavenProject` 
3. Call `MavenMojoProjectParser.generateProvenance(mavenProject)`

 
## What did you expect to see?
I'd expect `java.specification.version` to be used instead of `java.runtime.version` 
Then the `JavaVersion` marker has a `sourceCompatibility` and `targetCompatibility` of `17` (in my case)

## What did you see instead?
 The `JavaVersion` marker has a `sourceCompatibility` and `targetCompatibility` set as the full version, e.g. `17.0.2+8-LTS`

## What is the full stack trace of any errors you encountered?
-

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?
I'll provide a PR.

fabapp2:
See https://github.com/openrewrite/rewrite-maven-plugin/pull/594

fabapp2:
Maybe misunderstood the intent of API, `majorRealeaseVersion` and `majorTargetVersion` having `java.specification.value`.
I'd still assume `targetCompatibility` and `sourceCompatibility` should contain the specification values as they would be used in Maven because of their similarity in name to the Maven concept of target and source compatibility. 
However, if your intent was a different one, just close the issue and the PR.

knutwannheden:
@fabapp2 This sounds reasonable. Thanks for reporting and the PR!

PR #594, opened 2023-07-04
 Use java.specification.version when no source and target version exists

fabapp2:
## What's changed?
MavenMojoProjectParser uses `java.specification.version` as default for source and target version.

## What's your motivation?
"Fix": https://github.com/openrewrite/rewrite-maven-plugin/issues/593

### Checklist
- [x] I've added unit tests to cover ~~both~~the positive ~~and negative~~ cases, no dedicated tests existed before
- [x] I've added the license header to any new files through ~~`./gradlew licenseFormat`~~ copied from RewriteRunMojo. Gradle not used in this Module and most classes don't have a license header.
- [x] I've used the IntelliJ auto-formatter on affected files, it formatted only previously existing code so I didn't commit this change

### DIFF
--- a/src/test/java/org/openrewrite/maven/MavenMojoProjectParserTest.java
+++ b/src/test/java/org/openrewrite/maven/MavenMojoProjectParserTest.java
@@ -1,0 +1,34 @@
+package org.openrewrite.maven;
+
+import it.unimi.dsi.fastutil.ints.IntSets;
+import org.apache.maven.execution.MavenSession;
+import org.apache.maven.plugin.logging.SystemStreamLog;
+import org.apache.maven.project.MavenProject;
+import org.apache.maven.rtinfo.internal.DefaultRuntimeInformation;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.openrewrite.java.marker.JavaVersion;
+import org.openrewrite.marker.Marker;
+
+import java.util.Collection;
+import java.util.Collections;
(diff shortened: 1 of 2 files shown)


src/test/java/org/openrewrite/maven/MavenMojoProjectParserTest.java
```diff
@@ -0,0 +1,34 @@
+package org.openrewrite.maven;
```
knutwannheden, line 1: This file is lacking the license header. Use `./gradlew lF` to add it (or manually copy it from somewhere).

timtebeek: It looks like we rarely apply license headers in this project; likely because it's a Maven not a Gradle project. There are however license headers on some `package-info.java` files, such as for this package, albeit in src/main: https://github.com/openrewrite/rewrite-maven-plugin/blob/main/src/main/java/org/openrewrite/maven/package-info.java

I'd suggest we not require that change for this PR just yet, and instead separately tackle the issue of whether we want to header license headers to each file in Maven projects or not. The way things are going we'll likely automate that change through code review comments soon anyway, so maybe it's not necessary right now.

knutwannheden: Yes, of course a `./gradlew lF` won't help in this repo...

I also think we should tackle this independently of this PR. We should add a plugin to make sure that we have license headers for all of our source files.

PR #596, opened 2023-07-07
Add license headers and `license-maven-plugin`

knutwannheden:
The `license-maven-plugin` has not yet been configured to always run in builds, but that is the plan.


timtebeek (approved):

timtebeek:
Does this already include a gentle nudge when you're lacking the correct header? Or would we want to tackle that separately, ideally with the CLI?

knutwannheden:
So far I haven't configured the plugin to run in any of the CI builds. I still want to explore the options there.

Issue #600, opened 2023-07-18
Order of parsed Maven modules is sometimes not correct


fabapp2:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.1.6
- Maven plugin v5.2.6

## What is the smallest, simplest way to reproduce the problem?

This test 

https://github.com/spring-projects-experimental/spring-boot-migrator/blob/spike/rewrite-maven-parser/sbm-rewrite-maven-parser/src/test/java/org/springframework/sbm/parsers/RewriteMavenProjectParserIntegrationTest.java

should fail because `module-a/pom.xml` should be parsed last as it depends on `module-b/pom.xml`

## Are you interested in [contributing a fix to OpenRewrite]
I will provide a PR.

fabapp2:
I fixed the link to the test, @timtebeek

timtebeek:
This one is starting to make more sense after seeing the work done in
- https://github.com/openrewrite/rewrite-maven-plugin/pull/720

I don't think the work there will have fixed this issue too, but it does help me understand and validate the work you've done in
- https://github.com/openrewrite/rewrite-maven-plugin/pull/601

timtebeek:
@philippe-granet did you ever come across the problem outlined in this issue and the attached pull request #601 ? Seems like something that could have affected you too, especially when running in parallel, so wondering if you have any insight there.

philippe-granet:
What could be the impacts if the poms are loaded out of order?

timtebeek:
I'm not sure what exact problem Fabian had; do you recall @fabapp2?

PR #601, opened 2023-07-18
Order of parsed Maven modules is sometimes not correct

fabapp2:
## What's changed?
The method `MavenMojoProjectParser#collectPoms` adds the poms in the wrong order.
I think the problem can be fixed by using `MavenSession#getProjects()` instead.

## What's your motivation?
- #600 

I need some advice on how you think this should be tested.

My preferred approach would have been to test the method `collectPoms` and verify that the poms added to the `paths` exactly in the order as returned by `MavenSession#getProjects()`.
I would have liked to pass in a Mock for `MavenSession` and verify the call.
But there are some problems with this approach:
- The method is private, I would have been nasty and used reflection to make it accessible.
- No Mockito is available

And an additional test that verifies the expected behavior of `MavenSession#getProjects()`- the correct ordering

The next (additional) way to test this I'd see is an integration test for the plugin itself and verifying the parsing events. I did not understand `RewriteRunIT` and would be interested in your opinion of how to approach this  before proceeding.

@timtebeek 

By decoupling the `MavenMojoProject` parser from `File` and Maven unittest would become easier. 

### Checklist
- [ ] I've used the IntelliJ auto-formatter on affected files

### DIFF
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -193,7 +193,7 @@ public class MavenMojoProjectParser {
 
     public List<Marker> generateProvenance(MavenProject mavenProject) {
 
-        String javaRuntimeVersion = System.getProperty("java.runtime.version");
+        String javaRuntimeVersion = System.getProperty("java.specification.version");
         String javaVendor = System.getProperty("java.vm.vendor");
 
         String sourceCompatibility = null;
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -385,11 +385,9 @@ public class MavenMojoProjectParser {
         MavenProject topLevelProject = mavenSession.getTopLevelProject();
         logInfo(topLevelProject, "Resolving Poms...");
 
-        Set<Path> allPoms = new LinkedHashSet<>();
-        mavenProjects.forEach(p -> collectPoms(p, allPoms));
-        for (MavenProject mavenProject : mavenProjects) {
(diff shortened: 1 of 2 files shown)


timtebeek:
I've struggled a bit locally to see if we can test this through the integration test feature, but no luck thus far. My breakpoints aren't triggered, and even log output I add isn't showing; Perhaps you'd have more luck; [my branch is here](https://github.com/openrewrite/rewrite-maven-plugin/compare/tim/test_multi_module_order).

I'm exploring a bit the logic in SBM and the modules going into the Maven parser, roughly starting here:
https://github.com/spring-projects-experimental/spring-boot-migrator/blob/fcdfe7134e07386f255d288a6fd0dc37e98ad36b/sbm-rewrite-maven-parser/src/main/java/org/springframework/sbm/parsers/RewriteMavenProjectParser.java#L148-L153
There's a TODO comment there regarding the order, but since `runPerSubmodule` is false I don't think the first `mavenProject` argument to `listSourceFiles` is even being used there.

Instead we look to consistently use `mavenSession.getProjects()` to iterate over the modules in order returned by the session:
https://github.com/openrewrite/rewrite-maven-plugin/blob/9ffe2b33e723226c1843dd9df2f44fde2ee24dd6/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L138-L141
Not quite sure how that might end up with a different order as compared to your changes here. 🤔 

Will continue later, unless you have any new insights given the above.

fabapp2:
I understand that's a tricky part. I would think the problem will be solved with 9ffe2b3.
If you want we could see if we can use something from [here](https://github.com/spring-projects-experimental/spring-boot-migrator/tree/spike/rewrite-maven-parser/sbm-support-rewrite/src/test/java/org/springframework/sbm/parsers). 
There we create a Plexus container, call Maven and use a listener to get access to the session which can then be used to feed tests. Not straightforward but it allows testing methods depending on Maven components (that are difficult to initialize).

PR #615, opened 2023-08-08
Dump invalid message configuration

ammachado:
## What's changed?

Dump invalid recipes if maven debug log is enabled

## What's your motivation?

Try to identify invalid recipes in a large refactor project

### Checklist
- [ ] I've used the IntelliJ auto-formatter on affected files

### DIFF
--- a/src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
+++ b/src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
@@ -226,9 +226,14 @@ public abstract class AbstractRewriteMojo extends ConfigurableRewriteMojo {
             List<Validated.Invalid<Object>> failedValidations = validations.stream().map(Validated::failures)
                     .flatMap(Collection::stream).collect(toList());
             if (!failedValidations.isEmpty()) {
-                failedValidations.forEach(failedValidation -> getLog().error(
-                        "Recipe validation error in " + failedValidation.getProperty() + ": " +
-                        failedValidation.getMessage(), failedValidation.getException()));
+                failedValidations.forEach(failedValidation -> {
+                    if (getLog().isDebugEnabled()) {
+                        getLog().debug("Recipe " + recipe + " invalid");
+                    }
+                    getLog().error(
+                            "Recipe validation error in " + failedValidation.getProperty() + ": " +
+                                    failedValidation.getMessage(), failedValidation.getException());
+                });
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
```diff
@@ -226,9 +226,14 @@ protected ResultsContainer listResults(ExecutionContext ctx) throws MojoExecutio
...
-                failedValidations.forEach(failedValidation -> getLog().error(
-                        "Recipe validation error in " + failedValidation.getProperty() + ": " +
-                        failedValidation.getMessage(), failedValidation.getException()));
+                failedValidations.forEach(failedValidation -> {
+                    if (getLog().isDebugEnabled()) {
+                        getLog().debug("Recipe " + recipe + " invalid");
+                    }
+                    getLog().error(
+                            "Recipe validation error in " + failedValidation.getProperty() + ": " +
+                                    failedValidation.getMessage(), failedValidation.getException());
```
timtebeek, line 235: I wonder if instead we should qualify the property with the recipe name prefix here; since running with debug mode enabled might be uncommon. What are your thoughts @joanvr ?

timtebeek:
Thanks for your proposal @ammachado ! I've looked into this locally; I'm having a bit of trouble recreating when this would produce additional informative output. When I for instance intentionally misconfigure a recipe in `rewrite.yml`
```yaml
type: specs.openrewrite.org/v1beta/recipe
name: org.example.Recipe
displayName: Example
description: Example.
recipeList:
  - org.openrewrite.text.FindAndReplace:
      regex: maybe
```
Then I already get validation errors such as:
```
[ERROR] Recipe validation error in org.openrewrite.text.FindAndReplace: Unable to load Recipe: java.lang.IllegalArgumentException: Cannot deserialize value of type `java.lang.Boolean` from String "maybe": only "true" or "false" recognized
 at [Source: UNKNOWN; byte offset: #UNKNOWN] (through reference chain: org.openrewrite.text.FindAndReplace["regex"])
```
Do you recall an example where the recipe name was not included with the `failedValidation.getProperty()` we already print?

timtebeek:
As much as we appreciate the effort to improve the experience for validation issues, I'm not sure when this change would produce any additional information in the current form. My experience locally already provided the information needed to troubleshoot a misconfiguration, so unless we get a better understanding of cases not yet covered I propose to close this issue to clear the backlog.

PR #616, opened 2023-08-10
Log parsed files to Maven debug log

### LINKED ISSUE #544: Honor maven debug mode to log progress
In Maven, we can run in debug mode running `-X` 

```
 -X,--debug                             Produce execution debug output
```

Currently, the plugin does not honor this parameter, and it is unclear what is needed to enable debug logs

Ideally, I'd like to run this: 

```shell
 mvn -X -U org.openrewrite.maven:rewrite-maven-plugin:run Drewrite.activeRecipes=org.openrewrite.maven.OrderPomElements
```

And get all the internal `debug` logs of this plugin

timtebeek:
## What's changed?
We now debug log parsed files through the progress bar interface added to rewrite-all, such that we can optionally swap in a different 

## What's your motivation?
Fixes #544

Right now rewrite-all has runtime only dependencies; do we expect that to stay that way?
If those runtime dependencies change into transitive dependencies that could increase the cumulative runtime size here.

This minimal implementation for now only prints parsed files when running the OpenRewrite parsers, as Maven itself already logs the session, projects and mojos. No need to duplicate that here.

### DIFF
--- a/src/main/java/org/openrewrite/maven/ux/ProgressBarParsingEventListener.java
+++ b/src/main/java/org/openrewrite/maven/ux/ProgressBarParsingEventListener.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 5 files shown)


src/main/java/org/openrewrite/maven/ux/ProgressBarParsingEventListener.java
```diff
@@ -0,0 +1,35 @@
...
+
+    private final ProgressBar progressBar;
+
+    public ProgressBarParsingEventListener(ProgressBar progressBar) {
+        this.progressBar = progressBar;
+    }
+
+    @Override
+    public void parsed(Parser.Input input, SourceFile sourceFile) {
+        progressBar.intermediateResult("Parsed " + input.getPath());
```
timtebeek, line 33: Would we want to add any details here based on the produced SourceFile or ParseError?

jkschneider: This is incorrect anyway, as it is called _after_ the file is parsed.

jkschneider (changes requested):

timtebeek:
Closed as requested while we explore alternatives to report progress instead of logging.

Issue #538, opened 2023-04-05
Add KotlinParser to the maven plugin.


traceyyoshima:
The majority of Kotlin sources in the default repository list as of 4/5/23 are from Maven projects which causes the files to be `Quarks`.

timtebeek:
Now likely handled through https://github.com/search?q=repo%3Aopenrewrite%2Frewrite-all%20kotlin&type=code , which should make it's way here.

traceyyoshima:
Closing as a duplicate of https://github.com/openrewrite/rewrite-maven-plugin/issues/526.
We found that most of the Kotlin projects were Gradle.
However, many of the Gradle projects used the Android plugin.

Until recently, no one had asked for the KotlinParser in a maven project, so it hasn't been a priority.

Issue #526, opened 2023-03-13
Add support for parsing `Kotlin` sources


traceyyoshima:
This issue will track adding support for parsing `.kt` sources with the `KotlinParser` to the maven plugin.

jwillebrands:
Heya. Is there anything we can do to move forward on this? Our company unfortunately chose Maven for both Java and Kotlin projects and with a growing Kotlin codebase would benefit greatly from this support. Happy to help out if need be.

Wallman:
@jwillebrands This seems to be coming soon :) 

https://rewriteoss.slack.com/archives/C01A843MWG5/p1695762656099699?thread_ts=1695371865.357389&channel=C01A843MWG5&message_ts=1695762656.099699

Wallman:
Seems to be solved with https://github.com/openrewrite/rewrite-maven-plugin/pull/641

knutwannheden:
Please note that this is not yet part of any release, but that should follow this or next week.

knutwannheden:
I think it will come with the next patch release. You can see this by checking the commits since that release: https://github.com/openrewrite/rewrite-maven-plugin/compare/v5.8.1...main

knutwannheden:
Although in the 5.9.0-SNAPSHOT build it should be included.

Wallman:
https://github.com/openrewrite/rewrite-maven-plugin/releases/tag/v5.9.1

PR #619, opened 2023-08-21
Add kotlin autoDetect

kunli2:
Add Kotlin `autoDetect` to maven plug-in.


knutwannheden:
@kunli2 As far as I remember this also requires some corresponding changes in `moderne-ast-write`. Can you look into that and create a corresponding PR?

pom.xml
```diff
@@ -82,6 +82,7 @@
         <!-- Pinned versions, as RELEASE would make it into the published pom.xml -->
         <rewrite.version>8.6.0-SNAPSHOT</rewrite.version>
         <rewrite.python.version>1.2.0-SNAPSHOT</rewrite.python.version>
+        <rewrite.kotlin.version>1.4.0-SNAPSHOT</rewrite.kotlin.version>
```
timtebeek, line 85:
```suggestion
        <rewrite.kotlin.version>1.5.0-SNAPSHOT</rewrite.kotlin.version>
```

timtebeek:
Not quite sure what the current status is on this PR, or why we've held off on
- #526 
- #538 

Would we want to add the KotlinParser here directly?

Or do we plan to move to [the OmniParser in rewrite-polyglot](https://github.com/openrewrite/rewrite-polyglot/blob/v1.1.1/src/main/java/org/openrewrite/polyglot/OmniParser.java)?

knutwannheden:
@kunli2 Meanwhile I think we have the Kotlin parser integrated in the Maven plugin, so I think this is something we can merge, right?

PR #624, opened 2023-09-09
Update default plain text mask to include .config and markdown files

jonesbusy:
## What's changed?

The default list of plain text mask

## What's your motivation?

Rewriting mardown files and .config extension (for example `.mvn/maven.config` or `.mvn/jvm.config`)

Directly override the `plainTextMasks` on the plugin config. But I think having those default is a valid use case.

### Checklist
- [x] I've updated the documentation (if applicable)

Doc is : https://github.com/openrewrite/rewrite-docs/pull/231

Also the list was not up-to-date


src/main/java/org/openrewrite/maven/ConfigurableRewriteMojo.java
```diff
@@ -133,6 +134,7 @@ protected Set<String> getPlainTextMasks() {
                     "**/*.jsp",
                     "**/*.ksh",
                     "**/lombok.config",
```
timtebeek, line 136:
```suggestion
```

timtebeek:
Thanks for the suggestion! Should become available in the next release. You can try this out through our latest snapshots once we resolve the build issues on the main branch.

PR #629, opened 2023-09-25
Pass MavenExecutionContextView into MavenParser

timtebeek:
This view containing the Maven settings above is only created, not used or passed down.


timtebeek:
We call `.putMessage` on the delegate all the same, so looks like we don't have to pass this down. It's puzzling a bit, but that's about it.

Issue #630, opened 2023-09-26
Sort output of discover


Bananeweizen:
## What problem are you trying to solve?
Make the output of the "discover" goal more easy to understand. Right now it's only partially sorted. Therefore composite recipes are often NOT near their contained recipes. Active recipes are completely shuffled, which makes it hard to find any active recipe in the list.

## Describe the solution you'd like
* Sort the output lexicographically. That means all the styles and recipes under the different categories.
* Unify the indentation. Styles have way less (and better!) indentation than recipes.

See arrows in the screenshot showing the missing sort order. 
![grafik](https://github.com/openrewrite/rewrite-maven-plugin/assets/406876/c81ea6e6-d965-409c-9ffe-89f30a2184a4)

timtebeek:
Looks like the repeat here is off; it likely should just contain a single space.
https://github.com/openrewrite/rewrite-maven-plugin/blob/18a22a1ce7a9d1afa2db255994de535c098e6ce2/src/main/java/org/openrewrite/maven/RewriteDiscoverMojo.java#L126-L127

Would you be open to try that out and create a pull request?

PR #631, opened 2023-09-27
Fix indentation of available recipes output (#630)

Bananeweizen:
## What's changed?
Indentation of discover mojo is now as expected.

grafik

## What's your motivation?
Part of #630.


timtebeek (approved):

timtebeek:
Thanks a lot! That solves one part of #630. The sorting I imagine to be close trip the fix here.

PR #632, opened 2023-09-27
Sort output of discover mojo (#630)

### LINKED ISSUE #630: Sort output of discover

Bananeweizen:
* sort just for output, not in any of the APIs returning the data
* sort case insensitive, which is generally better suited for humans reading a list

## What's changed?
Sort the lists produced by the discover mojo output for easier reading by humans.

## What's your motivation?
Fixes #630

Looks like this now: 
grafik


timtebeek (approved):
Thanks a lot for this fix too!

PR #641, opened 2023-10-09
Add kotlin parser

kunli2:
Add kotlin parser to support Maven projects.

Verified in https://github.com/vert-x3/vertx-client-services and running recipe `org.openrewrite.kotlin.format.AutoFormat` and `org.openrewrite.java.ChangeType`, it works well.

### DIFF
--- a/src/main/java/org/openrewrite/maven/ResourceParser.java
+++ b/src/main/java/org/openrewrite/maven/ResourceParser.java
@@ -55,11 +55,14 @@ public class ResourceParser {
      */
     private final JavaParser.Builder<? extends JavaParser, ?> javaParserBuilder;
 
+    private final KotlinParser.Builder kotlinParserBuilder;
+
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/maven/ResourceParser.java
```diff
@@ -214,6 +220,11 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
             alreadyParsed.addAll(pythonPaths);
         }
 
+        if (!kotlinPaths.isEmpty()) {
+            sourceFiles = Stream.concat(sourceFiles, (Stream<S>) kotlinParser.parse(kotlinPaths, baseDir, ctx));
```
traceyyoshima, line 224: This is a great start, but it requires a few changes.

The JavaParser in the ResourceParser is meant to parse Java sources that haven't been parsed in main or test.
These sources do not have the classpath setup and will be missing type attribution.

The main and test Java sources set the classpath like [this](https://github.com/openrewrite/rewrite-maven-plugin/blob/main/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L310).

Issue #662, opened 2023-11-21
Add support for parsing Groovy sources (incl. Jenkinsfiles)


simonzn:
## What problem are you trying to solve?

We use Maven plugins for all CI tasks and would also like to run rewrite recipes for updating Jenkinsfiles through this plugin. 

## Describe the solution you'd like

The [GroovyParser](https://github.com/openrewrite/rewrite/blob/main/rewrite-groovy/src/main/java/org/openrewrite/groovy/GroovyParser.java) is already able to parse Jenkinsfiles, and I have tested it successfully with our pipelines. We would like the Maven plugin to parse Jenkinsfiles as Groovy sources by default instead of Quarks.

Currently the name must be "Jenkinsfile":
```
    @Override
    public boolean accept(Path path) {
        return path.toString().endsWith(".groovy") ||
               path.toFile().getName().equals("Jenkinsfile");
    }
``` 

We have two pipelines in each repository, "Jenkinsfile" for building the project and "Jenkinsfile_Release" for releasing it. It is quite common to have some suffix for Jenkinsfiles, so we would like the GroovyParser to accept `Jenkinsfile*` (or maybe even `*Jenkinsfile*` to support prefixes as well - although we don't use them).
 
## Have you considered any alternatives or workarounds?

Currently we execute text recipes on Jenkinsfiles by passing `-Drewrite.plainTextMasks="**/Jenkinsfile*,...` (otherwise they are only added as Quarks). This works for us for now, but may become limiting. Adding the GroovyParser would also mean we could use text recipes without modifying the plainTextMasks.

timtebeek:
Thanks for the reminder/suggestion @simonzn ! I've added a first commit to parse any file that starts with Jenkinsfile as Groovy in https://github.com/openrewrite/rewrite/commit/7ed7222b17f689d57c9392fd6430f82fe301da1b.

The logic for which parsers to use in the rewrite-maven-plugin is here. https://github.com/openrewrite/rewrite-maven-plugin/blob/558223d80e56302d4ee1dc939653d5dec19383f3/src/main/java/org/openrewrite/maven/ResourceParser.java#L135-L247

Would you be open to adding the GroovyParser in there as well? Even a draft PR would be fine to get this going.

simonzn:
Thanks for your swift reply @timtebeek ! I meant to do that, but then I found #641 and realized there might be more to it.

[...]

PR #663, opened 2023-11-21
Adds GroovyParser (#662)

simonzn:
Adds GroovyParser to the ResourceParser so we get source files with content for Groovy sources and Jenkinsfiles, instead of Quarks.

Fixes https://github.com/openrewrite/rewrite-maven-plugin/issues/662

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -206,6 +206,10 @@
         </dependency>
         <dependency>
             <groupId>org.openrewrite</groupId>
+            <artifactId>rewrite-groovy</artifactId>
+        </dependency>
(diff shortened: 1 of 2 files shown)


timtebeek (approved):
Looks good to me; indeed good enough for now not to add the classpath to the builder; the goal here is to parse Jenkinsfiles. Not yet to parse projects using `src/main/groovy` with a classpath.

timtebeek:
Confirmed to work for a petclinic migration as well; I'd say good to go. Thanks for the quick work!

PR #660, opened 2023-11-16
Avoid calling parsers when there are no source files

Bananeweizen:
With empty lists of main or test, Java or Kotlin files avoid invoking the parsers completely. That's more performant and avoids any bad side effects during initialization of the parsers.

Fixes #658.

## Anything in particular you'd like reviewers to focus on?
You may want to reformat, I used the Eclipse formatter.

### Checklist
- [ ] I've added unit tests to cover both positive and negative cases
- [ ] I've added the license header to any new files through `./gradlew licenseFormat`
- [ ] I've used the IntelliJ IDEA auto-formatter on affected files

PR #665, opened 2023-11-25
Simplify logic surrounding building of Java & Kotlin parsers.

timtebeek:
## What's changed?
- Simplify logic surrounding building Java & Kotlin parsers when there are no source files to parse.
- Also simplify logic related to log statements, markers & return value.
- https://github.com/openrewrite/rewrite-maven-plugin/pull/660

## What's your motivation?
Readability suffered a bit when mentally trying to parse what we had before.
```java
Stream<? extends SourceFile> cus = mainJavaSources.isEmpty() ? Stream.empty()
        : Stream.of(javaParserBuilder).map(JavaParser.Builder::build)
                .flatMap(parser -> parser.parse(mainJavaSources, baseDir, ctx));
```
Then further improvements from there on as spotted.

Could have merely gone for `javaParserBuilder.build()` in ternary, but that didn't yet help readability much.

### DIFF
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -323,36 +323,34 @@ public class MavenMojoProjectParser {
         javaParserBuilder.classpath(dependencies).typeCache(typeCache);
         kotlinParserBuilder.classpath(dependencies).typeCache(new JavaTypeCache());
 
-        Stream<? extends SourceFile> cus = mainJavaSources.isEmpty() ? Stream.empty()
-                : Stream.of(javaParserBuilder).map(JavaParser.Builder::build)
-                        .flatMap(parser -> parser.parse(mainJavaSources, baseDir, ctx));
+        Stream<SourceFile> parsedJava = Stream.empty();
+        if (!mainJavaSources.isEmpty()) {
+            parsedJava = javaParserBuilder.build().parse(mainJavaSources, baseDir, ctx);
+            logDebug(mavenProject, "Scanned " + mainJavaSources.size() + " java source files in main scope.");
+        }
 
-        Stream<? extends SourceFile> kcus = mainKotlinSources.isEmpty() ? Stream.empty()
-                : Stream.of(kotlinParserBuilder).map(KotlinParser.Builder::build)
-                        .flatMap(parser -> parser.parse(mainKotlinSources, baseDir, ctx));
+        Stream<SourceFile> parsedKotlin = Stream.empty();
+        if (!mainKotlinSources.isEmpty()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -323,36 +323,34 @@ private Stream<SourceFile> processMainSources(
...
-
-        Stream<SourceFile> parsedKotlin = kcus.map(addProvenance(baseDir, mainProjectProvenance, generatedSourcePaths));
-        logDebug(mavenProject, "Scanned " + mainKotlinSources.size() + " kotlin source files in main scope.");
-
         //Filter out any generated source files from the returned list, as we do not want to apply the recipe to the
         //generated files.
         Path buildDirectory = baseDir.relativize(Paths.get(mavenProject.getBuild().getDirectory()));
-        Stream<SourceFile> sourceFiles = Stream.concat(parsedJava, parsedKotlin).filter(s -> !s.getSourcePath().startsWith(buildDirectory));
+        Stream<SourceFile> sourceFiles = Stream.concat(parsedJava, parsedKotlin)
+                .filter(s -> !s.getSourcePath().startsWith(buildDirectory))
```
knutwannheden, line 345: Seems like it would be even better if we could filter out these generated sources before parsing them.

timtebeek: I had thought we parse them such that we have their types available for attribution, as indicated on line 302 above. What improvement would you make here then?

knutwannheden: I suppose we have to test this, but these sources should still not need to be parsed, because the type attribution in other sources is via class files.

timtebeek: I think it boils down to the question if `getCompileClasspathElements()` returns the classes compiled fom generated sources as well, which could depend on a number of different factors.
https://github.com/openrewrite/rewrite-maven-plugin/blob/0b8fc330f355a96b6e5b7a103601c07f775d9de3/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L317-L324

Right now we define generated sources as anything we find in the `target/` directory, and parse those as source rather than compiled classes.
https://github.com/openrewrite/rewrite-maven-plugin/blob/0b8fc330f355a96b6e5b7a103601c07f775d9de3/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L302-L308

While I agree it might be interesting to not parse those generated sources as source files but rather only as classpath entries, I'm not entirely sure that would work in all cases, and verifying that would take more time than I currently have available. Are you ok merging this PR as-is to at least simplify the logic?

knutwannheden (approved):

PR #672, opened 2023-12-02
Improve quark parsing message

### LINKED ISSUE #671: Have default excludes for known binary file types or disable maven resource folder parsing
Typical Maven projects contain resource folders, and these are part of the classpath, therefore they are currently scanned. Many of those files will not be handled by any recipe, therefore they should be excluded by default. Otherwise the log contains lots of lines about quarks parsing for included libraries, native executables, images etc. This is especially notable when setting the size threshold to 1 MB.

Default filters to contain *.jar, *.exe, *.png, *.gif, *.jpeg etc. 

Alternatively the resource folder could be ignored completely. In my naive assumption the parsing of the actual sources in the source folders should not be affected by that removal. However, someone with real insight into how rewrite works would need to assess that.

## Additional context

I have no idea how the core knows about file types that may or may not be handled by recipes in (custom) rewrite modules. If someone suddenly provided a PNG recipe (imagine a "remove geo tags from image" recipe), this would of course conflict with such default excludes.

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?

If it's just about default excludes, that might be possible.

Bananeweizen:
Fixes #671.

As suggested in the discussion, change the message to indicate that quark files are not parsed.

Also change the unit from "Mb" to "MB", since "Mb" indicates Megabit. Add a blank between numbers and unit for readability and to be consistent with output in other parts of the code.

## What's changed?

Only the log message which is emitted for large non parseable files.

## What's your motivation?

Clarifies the confusion described in #671.

### DIFF
--- a/src/main/java/org/openrewrite/maven/ResourceParser.java
+++ b/src/main/java/org/openrewrite/maven/ResourceParser.java
@@ -118,8 +118,8 @@ public class ResourceParser {
                 if (!attrs.isOther() && !attrs.isSymbolicLink() &&
                     !alreadyParsed.contains(file) && !isExcluded(file)) {
                     if (isOverSizeThreshold(attrs.size())) {
-                        logger.info("Parsing as quark " + file + " as its size " + attrs.size() / (1024L * 1024L) +
-                                    "Mb exceeds size threshold " + sizeThresholdMb + "Mb");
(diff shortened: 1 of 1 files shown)


timtebeek (approved):
Thanks for picking this up, and the MB fix to go with it!

PR #685, opened 2023-12-13
Kotlin source dir parsing bugfix

nmck257:
Fix bug where Kotlin test source parsing actually parsed main dir a second time, and test sources were later parsed as additional resources

### Checklist


timtebeek (approved):
Thanks!

PR #690, opened 2023-12-18
Revert to Java 17 after openrewrite/rewrite#3826

timtebeek:
## What's changed?
Revert back to Java 17 instead of Java 21.

## What's your motivation?
Fixes https://github.com/openrewrite/rewrite/issues/3826

## Any additional context
We saw users report failures running on Java 8 due to `MalformedParametersException: Invalid parameter name ""` after switching from https://github.com/openrewrite/rewrite-maven-plugin/compare/v5.14.1...v5.15.0
- https://github.com/openrewrite/rewrite/issues/3826

### DIFF
--- a/.github/workflows/bump-snapshots.yml
+++ b/.github/workflows/bump-snapshots.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/setup-java@v4
         with:
           distribution: temurin
-          java-version: 21
+          java-version: 17
           cache: maven
           server-id: ossrh
           settings-path: ${{ github.workspace }}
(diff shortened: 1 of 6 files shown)


timtebeek:
https://github.com/openrewrite/rewrite/issues/3826#issuecomment-1861028708

>  Spent a good deal of time wrestling with the debugger; turns out when built with Java 21 the names of the arguments going into `AbstractRewriteMojo_ResultsContainer_1_PlainTextVisitor` aren't retained, leading to the above Invalid parameter name "" deep down in java.lang.reflect.Executable.verifyParameters. Don't quite know what causes that.
> 
> So the problem is limited to Java 8 and the rewrite-maven-plugin, and only occurs when the rewrite-maven-plugin is built with Java 21 yet run on 8. As the Maven plugin will remain on 8 for the foreseeable future, and only the tests there use 17, there's no immediate rush in getting this fixed some other way.

PR #712, opened 2024-01-16
Deprecate mojo's property "activeRecipes" in favor of "rewrite.activeRecipes"

ghusta:
Deprecate mojo's property `activeRecipes` as it seems to be replaced by `rewrite.activeRecipes`.

Used in Maven goal `rewrite:run`.

Deprecated arguments can be displayed with help : `mvn org.openrewrite.maven:rewrite-maven-plugin:help -Dgoal=run -Ddetail`


timtebeek:
I like the initiative here @ghusta ! What I'm not quite sure about is how the interplay is between command line arguments, which use the `rewrite` prefix, and plugin `<configuration>` arguments, which take a list or set. Do you have any insights to share there?

We have the same pattern here for styles and exclusions. If we do decide to make a change it might be best to do so for all three.

ghusta:
Thank you @timtebeek for your feedback.

After a closer look, I decided to rework my contribution. That's why I'm going to close this PR.
I will open a new PR related to the issue I just submitted (#722).

Issue #649, opened 2023-11-01
A type incompatibility occurred while executing org.openrewrite.maven:rewrite-maven-plugin:5.9.1:run


laeubi:
## What version of OpenRewrite are you using?

I am using

- Maven plugin v5.9.1

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a multi module project.
```xml
<plugins>
<plugin>
			  <groupId>org.openrewrite.maven</groupId>
				<artifactId>rewrite-maven-plugin</artifactId>
				<version>5.9.1</version>
	<configuration>
			<activeRecipes>
			  <recipe>org.eclipse.ui.PlatformUISetHelp</recipe>
			</activeRecipes>
	</configuration>
</plugin>
</plugins>
```

The project is hosted here:
https://github.com/eclipse-platform/eclipse.platform

## What is the smallest, simplest way to reproduce the problem?

I added the above snippet to the root pom.xml
I created a rewrite.yml like this:
```
---
type: specs.openrewrite.org/v1beta/recipe
name: org.eclipse.ui.PlatformUISetHelp
displayName: Replace calls of PlatformUI.getWorkbench().getHelpSystem() with PlatformUI.getWorkbench().getHelpSystem().setHelp(...) with e4 safe PlatformUI.setHelp(...) variant
recipeList:
  - org.openrewrite.java.SimplifyMethodChain:
      methodPatternChain: ['org.eclipse.ui.PlatformUI getWorkbench()', 'org.eclipse.ui.IWorkbench getHelpSystem()', 'org.eclipse.ui.help.IWorkbenchHelpSystem setHelp(..)']
      newMethodName: setHelp
      matchOverrides: false
```

then I go to `debug/org.eclipse.debug.ui` in the project and call `mvn rewrite:run`

## What is the full stack trace of any errors you encountered?

```
[INFO] --- rewrite:5.9.1:run (default-cli) @ org.eclipse.debug.ui ---
[INFO] Using active recipe(s) [org.eclipse.ui.PlatformUISetHelp]
[INFO] Using active styles(s) []
[INFO] Validating active recipes...
[INFO] Project [[bundle] Debug UI] Resolving Poms...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  21.543 s (Wall Clock)
[INFO] Finished at: 2023-11-01T17:27:01+01:00
[...]

PR #721, opened 2024-01-24
Only add BuildTool marker to Maven Wrapper and pom.xml files

timtebeek:
## What's your motivation?
Fixes https://github.com/openrewrite/rewrite-maven-plugin/issues/649

We saw the plugin over report changes made to files in
- https://github.com/openrewrite/rewrite/issues/3948


shanman190:
I'm not sure on this one at the moment given over on the Gradle side, we've been messing with the `GradleProject` marker which is equivalent to the `MavenResolutionResult` marker on the Maven side. I think doing this has much more significant outcomes than meets the eye.

We'll chat more on slack though.

timtebeek:
You're right, this needs some more thought.

Issue #722, opened 2024-01-28
Refactor the Maven plugin 'rewrite.activeRecipes' property


ghusta:
## TL;DR

The type of `rewriteActiveRecipes` property can be refactored to `Set<String>`.
Also, the property `activeRecipes` in class `ConfigurableRewriteMojo` should also be deprecated or deleted in favor of `rewriteActiveRecipes`, as both properties have the same role.

## Rationale

The property, defined in `ConfigurableRewriteMojo`, can be configured in differents ways : with Maven CLI, or within Maven POM in XML.

However, in `ConfigurableRewriteMojo`, there are 2 fields to manage it : `activeRecipes` and `rewriteActiveRecipes`, which could be simplified.

When printing the plugin help, both properties are displayed, which can be confusing. No one is marked deprecated.

> mvn help:describe -Dplugin=org.openrewrite.maven:rewrite-maven-plugin:5.21.0 -Dgoal=run -Ddetail

## Recommended use

Here are the expected ways to configure it (according to https://docs.openrewrite.org/reference/rewrite-maven-plugin) :

### CLI

> mvn rewrite:run -D**rewrite.activeRecipes**=an.example.Recipe0,an.example.Recipe1

### POM

```xml
<configuration>
    <activeRecipes>
        <recipe>org.openrewrite.java.Spring</recipe>
        <recipe>org.openrewrite.maven.BestPractices</recipe>
    </activeRecipes>
</configuration>
```

## Not recommended use

However, as there are 2 fields to manage this property, this is also possible at the moment 😕.
These options are possible, but should not be recommended.

### CLI

> mvn rewrite:run -D**activeRecipes**=an.example.Recipe0,an.example.Recipe1

### POM

```xml
<configuration>
    <rewriteActiveRecipes>
        <recipe>org.openrewrite.java.Spring</recipe>
        <recipe>org.openrewrite.maven.BestPractices</recipe>
    </rewriteActiveRecipes>
</configuration>
```

## Worst situation

We could even have both properties defined, which is confusing 🤯.

### CLI

> mvn rewrite:run -D**rewrite.activeRecipes**=an.example.Recipe0 -D**activeRecipes**=an.example.Recipe1

## Maven plugin working

Maven is able to manage Java collections in mojos. It can use Collection, List, Set, etc. to map properties values.
There are some corner cases to be aware of : handling of null values / superfluous spaces.
It means that somes properties still need to be sanitized in Mojo.

[...]

PR #724, opened 2024-01-30
Standardize on `-Drewrite.` prefixed properties on the commandline

ghusta:
Fixes: #722

### DIFF
--- a/src/main/java/org/openrewrite/maven/ConfigurableRewriteMojo.java
+++ b/src/main/java/org/openrewrite/maven/ConfigurableRewriteMojo.java
@@ -45,11 +45,11 @@ public abstract class ConfigurableRewriteMojo extends AbstractMojo {
      */    
     @Parameter(property = "activeRecipes")
     @Deprecated
-    protected Set<String> deprecatedActiveRecipes = Collections.emptySet();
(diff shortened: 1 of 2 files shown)


timtebeek:
There's a small change to the order of recipes as reported in the logs, but other than that this looks good. Thanks for getting this started!

ghusta:
OK test failing because I didn't use LinkedHashSet to preserve order of recipes...

timtebeek:
Thanks a lot @ghusta ! Took your initial changes as an opportunity to consistently deprecate all such duplicated properties, and apply some more clean ups throughout. Would appreciate it if you could review if I indeed understood your docs in #722 correctly, or if there was any reason you'd held off on changing the other properties too.

timtebeek (approved):
Thanks again @ghusta ! Should help prevent any confusion about what properties to use going forward.

ghusta:
Thank you for your great cooperation when working on this PR.

timtebeek:
Yes really liked how you provided all the context to your proposed changes, and we now nudge folks towards the prefixed properties without breaking current use.

Hope to see you contribute again, this was great to see & review. :)

Issue #703, opened 2024-01-04
Unresolvable dependency leading to hard to debug issues


maxime-michel:
I have OS X and `org.openrewrite.maven.ChangeParentPom` is working fine locally. On our CI agent where I'd like to make it run for a batch of projects, I've noticed:

- the change is never done (although other recipes do work)
- with version 5.5.2 and up until version 5.17.0, I am told there were changes (even though I can't see them)
- with version 5.17.0+ (and I suppose https://github.com/openrewrite/rewrite/pull/3822), there are no changes and the message below is no longer present:

```
[WARNING] Changes have been made to pom.xml by:
[WARNING]     info.magnolia.rewrite.UseExtensionsPoms
[WARNING]         org.openrewrite.maven.ChangeParentPom: {oldGroupId=info.magnolia.services.maven.poms, newGroupId=info.magnolia.maven.poms-extensions, oldArtifactId=magnolia-services-incubator-pom, newArtifactId=magnolia-parent-pom-extensions, newVersion=60-SNAPSHOT}
[WARNING] Please review and commit the results.
```
My rewrite.yml:

    type: specs.openrewrite.org/v1beta/recipe
    name: info.magnolia.rewrite.UseExtensionsPoms
    recipeList:
    - org.openrewrite.maven.ChangeParentPom:
        oldGroupId: info.magnolia.services.maven.poms
        newGroupId: info.magnolia.maven.poms-extensions
        oldArtifactId: magnolia-services-incubator-pom
        newArtifactId: magnolia-parent-pom-extensions
        newVersion: 60-SNAPSHOT

How I run it: `mvn org.openrewrite.maven:rewrite-maven-plugin:5.17.1:run -Drewrite.activeRecipes=info.magnolia.rewrite.UseExtensionsPoms`

CI agent has Maven 3.9.6 and JDK 11.0.21+9. I'm now hours into this and still completely puzzled, so I figured I'd create this, at least to signal the possible problem with the upstream change above.

Sample pom:

<?xml version="1.0" encoding="UTF-8"?>
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <parent>
[...]

PR #725, opened 2024-01-31
Adding a logging resolution event listener

ammachado:
## What's changed?
Added a logging `ResolutionListener` on `MavenMojoProjectParser`.

## What's your motivation?
Identify maven resolution issues (#703).

No

No

Current maven debug output is not helpful regarding the plugin dependency resolution errors.

## Any additional context
No

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/maven/MavenLoggingResolutionEventListener.java
+++ b/src/main/java/org/openrewrite/maven/MavenLoggingResolutionEventListener.java
@@ -17,39 +17,17 @@ package org.openrewrite.maven;
 
 import org.apache.maven.plugin.logging.Log;
 import org.openrewrite.internal.lang.Nullable;
-import org.openrewrite.maven.tree.GroupArtifactVersion;
-import org.openrewrite.maven.tree.ManagedDependency;
-import org.openrewrite.maven.tree.MavenRepository;
-import org.openrewrite.maven.tree.Pom;
-import org.openrewrite.maven.tree.ResolutionEventListener;
-import org.openrewrite.maven.tree.ResolvedDependency;
-import org.openrewrite.maven.tree.ResolvedGroupArtifactVersion;
-import org.openrewrite.maven.tree.ResolvedPom;
-import org.openrewrite.maven.tree.Scope;
+import org.openrewrite.maven.tree.*;
 
 import java.util.List;
 import java.util.Objects;
 
-public class MavenLoggingResolutionEventListener implements ResolutionEventListener {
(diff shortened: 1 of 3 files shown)


src/main/java/org/openrewrite/maven/MavenLoggingResolutionEventListener.java
```diff
@@ -0,0 +1,124 @@
+/*
+ * Copyright 2020 the original author or authors.
```
timtebeek, line 2:
```suggestion
 * Copyright 2024 the original author or authors.
```

src/main/java/org/openrewrite/maven/MavenLoggingResolutionEventListener.java
```diff
@@ -0,0 +1,124 @@
...
+    public void dependencyManagement(ManagedDependency dependencyManagement, Pom containing) {
+        if (logger.isDebugEnabled()) {
+            logger.debug("Resolved managed dependency " + dependencyManagement + " on " + pomContaining(containing));
+        }
+    }
+
+    @Override
+    public void repository(MavenRepository mavenRepository, @Nullable ResolvedPom containing) {
+        if (logger.isDebugEnabled()) {
+            logger.debug("Resolving maven repository " + mavenRepository + pomContaining(containing));
```
timtebeek, line 108: As compared to the others here we use `Resolving` instead of `Resolved`; was that intentional?

timtebeek (commented):
I like it! Can see how this would be helpful to users of the OSS plugin too.

src/main/java/org/openrewrite/maven/MavenLoggingResolutionEventListener.java
ammachado: It is. `resolved` is used twice in the code, both times before the repository resolution happen (https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java#L230 and https://github.com/openrewrite/rewrite/blob/main/rewrite-maven/src/main/java/org/openrewrite/maven/internal/MavenPomDownloader.java#L508).

timtebeek:
Let me know when you consider this ready to review; as it's still marked as a draft.

timtebeek (approved):
Great to see this @ammachado ; After trying it out I've minimized the logged messages a bit, as in particular the resolved properties made debug mode unusable. Rather than trace I figured only keep the most valuable ones related to download success/failure and failure to access repositories.

Hope you agree!

PR #720, opened 2024-01-22
Delaying execution to the end of multi-module project

philippe-granet:
## What's changed?
Delaying execution to the end of multi-module project

## What's your motivation?
Fix #719

### Checklist
- [ ] I've added unit tests to cover both positive and negative cases
- [ ] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [X] I've used the IntelliJ IDEA auto-formatter on affected files

PR #726, opened 2024-01-31
Delaying execution to the end of multi-module project with Maven parallel build

### LINKED ISSUE #719: Delaying execution to the end of multi-module project
## What version of OpenRewrite are you using?
I am using latests version

## How are you running OpenRewrite?
I am using the maven goal `rewrite:runNoFork`, and my project is a multi-module project.

## What is the smallest, simplest way to reproduce the problem?
Run `mvn -T 1C package rewrite:runNoFork` on a multi-module project, **with Maven parallel execution** (https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3)

## What did you expect to see?
Recipes executed at end of all module packaging

## What did you see instead?
Recipes executed before the end of all module packaging

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?
Yes, the PR that fix the bug: https://github.com/openrewrite/rewrite-maven-plugin/pull/720

philippe-granet:
## What's changed?
Since PR #720 merged, I made more tests on different projects, and there is some cases where recipes are still executed before the end of all module packaging when building with Maven parallel build .

I analyzed the implementation of the `installAtEnd` and `deployAtEnd` options of the `maven-install-plugin`:  
https://github.com/apache/maven-install-plugin/blob/master/src/main/java/org/apache/maven/plugins/install/InstallMojo.java  
and `maven-install-plugin` plugins:
https://github.com/apache/maven-deploy-plugin/blob/master/src/main/java/org/apache/maven/plugins/deploy/DeployMojo.java

It uses mojo Context to store "state markers", presence of marker means project was "processed".   

This MR is based on this implementation.   

## What's your motivation?
Fix #719

@timtebeek 

## Any additional context
I have added a `RewriteRunParallelIT` test, with a Maven project that have 2 modules `a` and `b`.
These 2 modules are build in parallel (`@MavenOption(value = MavenCLIOptions.THREADS, parameter = "2")`.
And I've added a sleep Antrun task in module `a`, in order to force module `a` finish building after module `b`.`

Without this MR, this test fail, because previous implementation:
`List<MavenProject> sortedProjects = mavenSession.getProjectDependencyGraph().getSortedProjects();`
always return `[a,b]` project list in this order, even if module `a` is longer to build than module `b`.

### Checklist
- [X] I've added unit tests to cover both positive and negative cases
- [X] I've read and applied the [recipe conventions and best practices](https://docs.openrewrite.org/authoring-recipes/recipe-conventions-and-best-practices)
- [X] I've used the IntelliJ IDEA auto-formatter on affected files

### DIFF
--- a/src/main/java/org/openrewrite/maven/AbstractRewriteDryRunMojo.java
+++ b/src/main/java/org/openrewrite/maven/AbstractRewriteDryRunMojo.java
@@ -49,12 +49,18 @@ public class AbstractRewriteDryRunMojo extends AbstractRewriteMojo {
     public void execute() throws MojoExecutionException {
         if (rewriteSkip) {
             getLog().info("Skipping execution");
+            putState(State.SKIPPED);
             return;
         }
+        putState(State.TO_BE_PROCESSED);
 
         // If the plugin is configured to run over all projects (at the end of the build) only proceed if the plugin
         // is being run on the last project.
-        if (!runPerSubmodule && !isLastProjectInReactor()) {
+        if (!runPerSubmodule && !allProjectsMarked()) {
+            getLog().info("REWRITE: Delaying execution to the end of multi-module project for "
+                + project.getGroupId() + ":"
+                + project.getArtifactId()+ ":"
+                + project.getVersion());
             return;
         }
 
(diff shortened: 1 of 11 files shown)


timtebeek (approved):
Thanks a lot for coming back with the detailed fix!

PR #729, opened 2024-02-03
Separate JavaVersion markers for src/main and src/test

timtebeek:
## What's changed?
Create separate JavaVersion markers for src/main and src/test, based on their relative properties.
Fall back to src/main for src/test when not explicitly set.

## What's your motivation?
Allows adopting Java 17 language features in tests when src/main still uses Java 8, as evaluated by HasJavaVersion & UsesJavaVersion.

Files like the Maven wrapper files and other additional resources outside of src/main and src/test now get the runtime Java version marker, as they do not explicitly declare their Java version.

We could continue to assign the wrong src/main JavaVersion to src/test files for projects that use a different version for tests. That might end up surprising folks though, and does not allow some recipes to make changes in src/test when src/main uses a lower version.

## Any additional context
Prompted by a question for a community member/customer.
Would also allow projects to safely adopt Java 17+ features in tests first, before moving to main.

### DIFF
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -198,12 +198,16 @@ public class MavenMojoProjectParser {
             return sourceFile;
         }).filter(Objects::nonNull);
 
+        // Use runtime Java version for additional resources, as they do not explicitly use a source/target version.
+        List<Marker> additionalResourcesProvenance = new ArrayList<>(projectProvenance);
+        additionalResourcesProvenance.add(getJavaVersionMarker(null, null));
+
         // Collect any additional files that were not parsed above.
         int sourcesParsedBefore = alreadyParsed.size();
         Stream<SourceFile> parsedResourceFiles;
         if (parseAdditionalResources) {
             parsedResourceFiles = rp.parse(mavenProject.getBasedir().toPath(), alreadyParsed)
-                    .map(addProvenance(baseDir, projectProvenance, null));
+                    .map(addProvenance(baseDir, additionalResourcesProvenance, null));
             logDebug(mavenProject, "Parsed " + (alreadyParsed.size() - sourcesParsedBefore) + " additional files found within the project.");
         } else {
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -198,12 +198,16 @@ public Stream<SourceFile> listSourceFiles(MavenProject mavenProject, @Nullable X
             return sourceFile;
         }).filter(Objects::nonNull);
 
+        // Use runtime Java version for additional resources, as they do not explicitly use a source/target version.
+        List<Marker> additionalResourcesProvenance = new ArrayList<>(projectProvenance);
+        additionalResourcesProvenance.add(getJavaVersionMarker(null, null));
```
timtebeek, line 203: Instead of the runtime Java version, we could also opt to go for the src/main JavaVersion marker here too; not correct, but could be less surprising. In any case I'd expect it to be rare that these resources even need a JavaVersion marker, so this is mostly in case anyone built something we didn't foresee here, as we were surprised before with markers affecting wrappers for instance.

timtebeek:
@fabapp2 I had to change a test you added, but I don't think this change would otherwise affect your use of `MavenMojoProjectParser`. Let me know if you have any concerns though!

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
shanman190: I don't think adding the `JavaVersion` marker outside of the source set is correct. As a parallel from the Gradle plugin, over there we don't add a `JavaVersion` marker to anything that is outside of the source set either.

timtebeek: Even better; I'd kept this out of caution, but if you also feel the markers should not be there I'll remove them.

timtebeek: Picked up in https://github.com/openrewrite/rewrite-maven-plugin/pull/729/commits/2a2d1f625843bd871b20b6ebc2e21d386159a58b

shanman190 (approved):

Issue #734, opened 2024-02-06
Mistake in detecting encoding of pom.xml


leveretconey:
## What version of OpenRewrite are you using?
I am using

- OpenRewrite 8.13.1
- rewrite-maven-plugin 5.18.0
- Maven 3.5.0

## How are you running OpenRewrite?

pom.xml of the project: 

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>org.example</groupId>
  <artifactId>debug</artifactId>
  <version>1.0-SNAPSHOT</version>
  <properties>
    <!-- UTF-8编码的中文-->
    <!-- ↑ Chinese characters in utf-8-->
    <project.build.sourceEncoding>gbk</project.build.sourceEncoding>
  </properties>
</project>

```

In a legacy application that java source files are using gbk encoding while pom.xml is using utf-8 encoding, OpenRewrite has recognized the file encoding of pom.xml as gbk. This leads to some problem in our scenerio.

After some debugging, I noticed that OpenRewrite reads `project.build.sourceEncoding` property from pom and sets it as the encoding of all files in this project. I think this behavior is not correct because this property is only applicable to source code files by Maven's design. 

## What did you expect to see?
pom.xml should be detected as using utf-8 encoding

## What did you see instead?
pom.xml is detected as using gbk encoding

## Are you interested in [contributing a fix to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#contributing-fixes)?
I won’t contribute to this because the code changes may be huge and involve multiple code repositories.

leveretconey:
Thanks for your reply @timtebeek !

[...]

PR #735, opened 2024-02-06
Clear `project.build.sourceEncoding` after parsing sources

timtebeek:
## What's your motivation?
- https://github.com/openrewrite/rewrite-maven-plugin/issues/734

Wasn't sure if this exactly captures the issue described in #734 ; from the description there this seemed a good first step.

@leveretconey

Not really sure we have a lot of other options here; but open to help explore.


leveretconey:
@timtebeek Thank you! I confirm that this patch does solve by issue.

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -198,6 +198,11 @@ public Stream<SourceFile> listSourceFiles(MavenProject mavenProject, @Nullable X
             return sourceFile;
         }).filter(Objects::nonNull);
 
+        // Clear the charset after parsing or sources is complete, as it may not be applicable to other sources.
```
knutwannheden, line 201: I am not convinced that this actually works. Since we are returning a stream of source files, the parsing doesn't actually occur until after the return from this method.

knutwannheden: Or is the value from the execution context already read before any parsing starts?

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -198,6 +198,11 @@ public Stream<SourceFile> listSourceFiles(MavenProject mavenProject, @Nullable X
...
+        if (mavenSourceEncoding != null) {
+            ParsingExecutionContextView.view(ctx).setCharset(null);
```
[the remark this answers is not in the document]
timtebeek: Specifically we're offending these guidelines for the predicate.
image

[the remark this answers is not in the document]
timtebeek: Went ahead and applied the above to make it easier to review.

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -169,14 +168,15 @@ public Stream<SourceFile> listSourceFiles(MavenProject mavenProject, @Nullable X
...
 
-        Object mavenSourceEncoding = mavenProject.getProperties().get("project.build.sourceEncoding");
-        if (mavenSourceEncoding != null) {
-            ParsingExecutionContextView.view(ctx).setCharset(Charset.forName(mavenSourceEncoding.toString()));
-        }
         JavaParser.Builder<? extends JavaParser, ?> javaParserBuilder = JavaParser.fromJavaVersion()
                 .styles(styles)
                 .logCompilationWarningsAndErrors(false);
 
+        Object mavenSourceEncoding = mavenProject.getProperties().get("project.build.sourceEncoding");
```
knutwannheden, line 175: Ideally, we would also accept the encoding from the compiler plugin configuration: https://www.baeldung.com/maven-encoding

timtebeek: done in https://github.com/openrewrite/rewrite-maven-plugin/pull/735/commits/175d711c6a1febed5e8734adfc11b31bfc60cfee , thanks!

knutwannheden: Were you able to check if the precedence is correct between the plugin configuration and the project property?

knutwannheden (approved):
Apart from the detail in my comment this LGTM.

PR #747, opened 2024-03-02
Avoid 2 Java level warnings of the compiler plugin

Bananeweizen:
## What's changed?
The build complains about incomplete Java level settings. Once in the main source compilation ([example from main branch build](https://github.com/openrewrite/rewrite-maven-plugin/actions/runs/8083030796/job/22085154131#step:4:30)), and again in the test source compilation (only if using a JDK > 17). The PR avoids both warnings.

```
[INFO] --- compiler:3.12.1:compile (default-compile) @ rewrite-maven-plugin ---
[INFO] Recompiling the module because of changed source code.
[INFO] Compiling 23 source files with javac [debug target 1.8] to target\classes
[WARNING] bootstrap class path not set in conjunction with -source 8                                      <---------------------
[WARNING] source value 8 is obsolete and will be removed in a future release
[WARNING] target value 8 is obsolete and will be removed in a future release
[WARNING] To suppress warnings about obsolete options, use -Xlint:-options.
[INFO] Annotation processing is enabled because one or more processors were found
...
[INFO] --- compiler:3.12.1:testCompile (default-testCompile) @ rewrite-maven-plugin ---
[INFO] Recompiling the module because of changed dependency.
[INFO] Compiling 10 source files with javac [debug target 17] to target\test-classes
[WARNING] system modules path not set in conjunction with -source 17                                      <---------------------
[INFO] Annotation processing is enabled because one or more processors were found

```

* use only the maven.compiler.x properties
* set release instead of source/target, if possible
* conditionally set release on JDKs >= 9
* remove Javadoc setting (it takes the compiler property as default)

The changes are compatible even with JDK8 because of the profile using the JDK version as activation condition, but the current code doesn't cleanly build *on* JDK8 (with and without my change). If building *on* JDK8 (instead of *for* JDK8) is still a requirement, then that needs to be fixed separately (there are compile errors around the sneakyThrows).

A second thing to look at is the previous use of the system property `java.version`. IMO that's confusing as hell, because every JRE has a built-in system property with the same name. I'd therefore strongly suggest to not re-introduce such a property for any reason.

### Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,10 +69,11 @@
         <developerConnectionUrl>scm:git:ssh://git@github.com/openrewrite/rewrite-maven-plugin.git
         </developerConnectionUrl>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <java.version>1.8</java.version>
 
-        <maven.compiler.testSource>17</maven.compiler.testSource>
-        <maven.compiler.testTarget>17</maven.compiler.testTarget>
+        <maven.compiler.source>8</maven.compiler.source>
+        <maven.compiler.target>${maven.compiler.source}</maven.compiler.target>
+        <!-- a profile in the profiles section sets the maven.compiler.release property on supported JREs -->
+        <maven.compiler.testRelease>17</maven.compiler.testRelease>
 
         <!-- dependencies and plugins -->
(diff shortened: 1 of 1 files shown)


timtebeek (approved):
Good to know this pattern and get rid of that warning; thanks a lot!

PR #751, opened 2024-03-08
Printable Recipe Datatables

ryan-hudson:
## What's changed?
Adds rewrite.exportDatatables property to Maven Plugin to allow Datatable information to be exported as a CSV. 
See the[ Rewrite-Core PR](https://github.com/openrewrite/rewrite/pull/4087) for more information. 

Note that the datatable output path is currently set to target/rewrite/datatables
These changes are dependent on the changes in [this rewrite-core pr](https://github.com/openrewrite/rewrite/pull/4087)

### DIFF
--- a/src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
+++ b/src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
@@ -1,0 +1,393 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.openrewrite.maven;
+
+import io.micrometer.core.instrument.Metrics;
+import org.apache.maven.artifact.DependencyResolutionRequiredException;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.codehaus.plexus.classworlds.realm.ClassRealm;
+import org.openrewrite.Cursor;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.LargeSourceSet;
+import org.openrewrite.PrintOutputCapture;
+import org.openrewrite.Recipe;
+import org.openrewrite.RecipeRun;
+import org.openrewrite.Result;
+import org.openrewrite.SourceFile;
+import org.openrewrite.Tree;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.Validated;
+import org.openrewrite.config.Environment;
+import org.openrewrite.config.RecipeDescriptor;
+import org.openrewrite.internal.InMemoryLargeSourceSet;
+import org.openrewrite.internal.ListUtils;
+import org.openrewrite.internal.lang.Nullable;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.kotlin.tree.K;
+import org.openrewrite.marker.Generated;
+import org.openrewrite.marker.Marker;
+import org.openrewrite.marker.Markers;
+import org.openrewrite.marker.Markup;
+import org.openrewrite.marker.SearchResult;
+import org.openrewrite.style.NamedStyles;
+import org.openrewrite.xml.tree.Xml;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.UnaryOperator;
+import java.util.stream.Stream;
+
+import static java.util.Collections.emptyList;
+import static java.util.stream.Collectors.joining;
+import static java.util.stream.Collectors.toList;
+
+public abstract class AbstractRewriteBaseRunMojo extends AbstractRewriteMojo {
+
+    @Parameter(property = "rewrite.exportDatatables", defaultValue = "false")
+    protected boolean exportDatatables;
+
+    /**
+     * Attempt to determine the root of the git repository for the given project.
+     * Many Gradle builds co-locate the build root with the git repository root, but that is not required.
+     * If no git repository can be located in any folder containing the build, the build root will be returned.
+     */
+    protected Path repositoryRoot() {
+        Path buildRoot = getBuildRoot();
+        Path maybeBaseDir = buildRoot;
+        while (maybeBaseDir != null && !Files.exists(maybeBaseDir.resolve(".git"))) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
```diff
@@ -262,15 +267,22 @@ protected LargeSourceSet loadSourceSet(Path repositoryRoot, Environment env, Exe
...
-                        return !source.getBefore().getMarkers().findFirst(Generated.class).isPresent();
-                    }
-                    return true;
-                })
-                .collect(toList());
+        RecipeRun recipeRun = recipe.run(sourceSet, ctx);
+
+        if (exportDatatables) {
+            String timestamp = LocalDateTime.now().format(DateTimeFormatter.ofPattern("yyyy-MM-dd_HH-mm-ss-SSS"));
+            Path datatableDirectoryPath = Paths.get("target/rewrite/", "datatables", timestamp);
```
ammachado, line 274: Can you use `Paths.get("target", "rewrite", "datatables", timestamp);` here instead? Might be safer for Windows users.

ryan-hudson: Yep, good catch. Corrected.

shanman190: So just to chime in, the Windows `FileSystem` implementation accepts both forward and back slashes without any issues. A *nix based `FileSystem` will only accept forward slashes, so either always using forward slashes, letting the filesystem choose (as illustrated by @ammachado), or using `File.separatorChar` are all valid solutions that work across OS boundaries.

ryan-hudson:
@timtebeek Is there anything else needed to get this merged?

timtebeek:
hi @ryan-hudson thanks for the work! I've been out for a couple weeks, with a bit more travel to go still. Perhaps @knutwannheden or @sambsnyd can have a look.

ryan-hudson:
@timtebeek Bumping this again for visibility. It would be great to have this reviewed and merged!

src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
```diff
@@ -67,6 +69,9 @@ public abstract class AbstractRewriteMojo extends ConfigurableRewriteMojo {
     @Parameter(property = "rewrite.resolvePropertiesInYaml", defaultValue = "true")
     protected boolean resolvePropertiesInYaml;
 
+    @Parameter(property = "rewrite.exportDatatables", defaultValue = "false")
+    protected boolean exportDatatables;
```
timtebeek, line 73: I understand why you added the property here, as this class contains the `runRecipe` method. But looking at the wider class hierarchy I worry that this option then also shows up in the documentation for the CycloneDxBomMojo and RewriteDiscoverMojo: https://openrewrite.github.io/rewrite-maven-plugin/plugin-info.html

image

Perhaps we can revisit the hierarchy to only show this option where applicable? Perhaps push it down into `AbstractRewriteRunMojo` ?

src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
```diff
@@ -262,15 +267,22 @@ protected LargeSourceSet loadSourceSet(Path repositoryRoot, Environment env, Exe
...
-                        return !source.getBefore().getMarkers().findFirst(Generated.class).isPresent();
-                    }
-                    return true;
-                })
-                .collect(toList());
+        RecipeRun recipeRun = recipe.run(sourceSet, ctx);
+
+        if (exportDatatables) {
+            String timestamp = LocalDateTime.now().format(DateTimeFormatter.ofPattern("yyyy-MM-dd_HH-mm-ss-SSS"));
+            Path datatableDirectoryPath = Paths.get("target", "rewrite", "datatables", timestamp);
```
timtebeek, line 274: I wonder if we should include the timestamp here, or if we should just use stable predictable paths that get overwritten. Any thoughts on that yourself?

src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
ryan-hudson: Ideally this datatable export feature would be available for both regular runs and dry runs, which was my main reasoning for positioning it in `AbstractRewriteMojo` class as it happens to be the first (and only) parent class shared between those. 

I understand your concern regarding the other two classes though. We could consider moving the parameter and corresponding logic down into the `AbstractRewriteRunMojo` and `AbstractRewriteDryRunMojo` individually, but that would lead to some unfortunate duplication.

src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
ryan-hudson: I went back and forth on this as well. I think keeping timestamps is valuable, it provides a bit more flexibility and prevents accidental overwriting in cases where someone runs two different recipes that happen to produce the same datatable (i.e. the RecipeRunStats datatable for example). It does add a bit more bulk, but it is also in the target directory which is pretty ephemeral.

src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
timtebeek: Yeah there's not really an immediately clear satisfactory way to handle this; duplication is unfortunate as well, but so is unused documented options. :/

src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
timtebeek: then indeed let's keep the timestamps. I feel like either option would get criticized anyway, no matter which we choose. ;)

src/main/java/org/openrewrite/maven/AbstractRewriteMojo.java
ryan-hudson: So, looking into this a little further, both the `AbstractRewriteRunMojo` and `AbstractRewriteDryRunMojo` would have to override the `runRecipe` method from `AbstractRewriteMojo`, which would result in the `AbstractRewriteMojo` `runRecipe` implementation being dead code, as no other subclasses use that method. Making that method abstract is also not ideal, as it would force the `CycloneDxBomMojo` and `RewriteDiscoverMojo` classes to implement it unnecessarily. 

The best solution I can think of is to create another new Mojo class that extends `AbstractRewriteMojo` that `AbstractRewriteRunMojo` and `AbstractRewriteDryRunMojo` both extend. We could move any functionality that's only used in the two run mojo's there (including the `runRecipe` method and exportDatatables parameter), while keeping the other mojo's the same. This would be a bigger change but would probably be an overall better architecture. 

Definitely interested to hear your thoughts on this, or if you have any other ideas/approaches.

timtebeek: That might indeed be the best way to go about it; but we have painted ourselves into the corner a bit on what to name that new class. 😅 How about `BaseRewriteRunMojo` ? Open to alternatives!

ryan-hudson: Alright, did that class refactor. Opted to go with `AbstractRewriteBaseRunMojo` as the name of the new class, but also open to alternative naming.

src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
```diff
@@ -0,0 +1,393 @@
+/*
+ * Copyright 2020 the original author or authors.
```
timtebeek, line 2:
```suggestion
 * Copyright 2024 the original author or authors.
```

timtebeek (commented):
Thanks for the rework! I'll have to review the moved changes still, but I imagine those were mostly a lift over right?

ryan-hudson:
> Thanks for the rework! I'll have to review the moved changes still, but I imagine those were mostly a lift over right?

Yep, purely copy paste work. No intentional changes to any of the moved methods.

timtebeek (approved):
Thanks a lot @ryan-hudson for the work both here and in 
- https://github.com/openrewrite/rewrite/pull/4087

Should be need to see what others will build with data tables.

/cc @mike-solomon data tables are now no longer unique to Moderne, so we might want to document how to use this in OpenRewrite as well, [as we have done for Moderne](https://docs.moderne.io/user-documentation/moderne-platform/getting-started/data-tables).

PR #768, opened 2024-04-07
Parse inline checkstyle config from maven-checkstyle-plugin

tobli:
## What's changed?
Inline checkstyle config from `maven-checkstyle-plugin` can now be parsed, in addition to reading file references from the `configLocation` tag.

## What's your motivation?
Support more of the various config options that exist for `maven-checkstyle-plugin`, to capture more style instructions that exist in existing codebases. 

I didn't think of an obvious positive test case, but I'm happy to take suggestions. For now, just look for the absence of parsing errors.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/maven/ConfigurableRewriteMojo.java
+++ b/src/main/java/org/openrewrite/maven/ConfigurableRewriteMojo.java
@@ -16,6 +16,8 @@
 package org.openrewrite.maven;
 
 import com.puppycrawl.tools.checkstyle.Checker;
+import java.io.IOException;
+import java.io.StringWriter;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.AbstractMojo;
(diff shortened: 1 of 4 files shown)


src/main/java/org/openrewrite/maven/ConfigurableRewriteMojo.java
```diff
@@ -278,6 +292,18 @@ protected List<NamedStyles> loadStyles(MavenProject project, Environment env) {
         return styles;
     }
 
+    private String toCheckStyleDocument(final Xpp3Dom dom) throws IOException {
+        StringWriter stringWriter = new StringWriter();
+
+        MXSerializer serializer = new MXSerializer();
+        serializer.setOutput(stringWriter);
+        serializer.docdecl(CHECKSTYLE_DOCTYPE);
```
tobli, line 300: Checkstyle needs a doctype to be present in order to parse a config xml.

src/main/java/org/openrewrite/maven/ConfigurableRewriteMojo.java
```diff
@@ -278,6 +292,18 @@ protected List<NamedStyles> loadStyles(MavenProject project, Environment env) {
         return styles;
     }
 
+    private String toCheckStyleDocument(final Xpp3Dom dom) throws IOException {
+        StringWriter stringWriter = new StringWriter();
+
+        MXSerializer serializer = new MXSerializer();
```
tobli, line 298: This was what I came up with using the xml classes within the Plexus libraries. Wish it could be a bit cleaner.

timtebeek (approved):
Thanks for working out what's required to parse inline config; nice to be able to cover this variant as well!

tobli:
> Thanks for working out what's required to parse inline config; nice to be able to cover this variant as well!

My follow up is parsing check style files from plugin dependency jars. That requires some more refactoring, but with this the diff is smaller. I can post a rough draft PR soon for early feedback.

PR #782, opened 2024-05-02
Estimate time saved

### LINKED ISSUE #781: Sum and show time saved from running recipes, similar to Gradle plugin
Show folks how much time is estimated to be saved through running recipes, to show appreciation to recipe authors and encourage more use.

The same as was done for the Gradle plugin, but now applied to the Maven plugin
- https://github.com/openrewrite/rewrite-gradle-plugin/pull/294

nicolasb29:
Fixes #781

## What's changed?

After log changes for each result, sum estimate time saved and log it.

## What's your motivation?

OpenRewrite is working fine and it is very helpful. To show the importance and the time saved, I think it's interesting to show the estimate time saved.

The same was gone for [Gradle plugin ](https://github.com/openrewrite/rewrite-gradle-plugin/pull/294)


timtebeek (approved):
Thanks a lot for adding this here as well! Much apppreciated.

PR #789, opened 2024-05-21
Use Maven enforcer for Java and Maven minimum versions

Bananeweizen:
## What's changed?

* Use Maven enforcer to clearly indicate the need for Java 17 during build, as the usage of Java 8 on the classpath may lead to confusion.
* Also enforce the same minimum Maven version as used in the Maven wrapper to ensure both mvn and mvnw create similar output.

## What's your motivation?

If one tries to build on a JDK less than 17, the error messages are rather confusing (because the maven compiler plugin complains about some unknown symbols only). Make the error more obvious by using Maven enforcer instead.

### Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -396,6 +396,29 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


timtebeek (approved):
Thanks a lot! Should indeed help prevent puzzling warnings when mismatched versions are used.

PR #800, opened 2024-06-16
Add support for artifact recipes as file

jonesbusy:
Hi,

The plugin already support `recipeArtifactCoordinates` that will resolve artifact from a remote repository and it's transitive dependency. I would like to support loading recipes directly from the file system without having the artifact published to a remote repository.

My use case imply that I don't want to add necessary maven repositories (via pom or settings) So I must resolve recipes locally.

I've just verified that tests are passing, but I didn't add any.

I've tested on a sample project

Right now `recipeArtifactCoordinates`takes precedence on `recipeArtifactFiles`when we can imagine support both by adding more classloader to the environment

```
mvn -Drewrite.activeRecipes=my.test.receipe.TestRecipe -Drewrite.recipeArtifactFiles=myjar.jar org.openrewrite.maven:rewrite-maven-plugin:5.34.0-SNAPSHOT:run
```

## What's changed?

See before

## What's your motivation?

See before

@timtebeek 

Yes by publishing to a remote temporary repository and updating maven repository. But is not convenient on my use case that I don't want to use any custom settings or update poms.

## Any additional context

The GSoC 2024 https://summerofcode.withgoogle.com/programs/2024/projects/anaMmWRR

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
+++ b/src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
@@ -85,10 +85,19 @@ public abstract class AbstractRewriteBaseRunMojo extends AbstractRewriteMojo {
             }
 
             URLClassLoader recipeArtifactCoordinatesClassloader = getRecipeArtifactCoordinatesClassloader();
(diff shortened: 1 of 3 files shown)


jonesbusy:
Closing the PR, at the end I can use the `recipeArtifactCoordinates` and get artifact from the local repository

Issue #799, opened 2024-06-14
Pass arguments from the command line into a single configurable recipe


commonquail:
We can already run OpenRewrite via the Maven plugin without modifying files up front:

```sh
mvn -U org.openrewrite.maven:rewrite-maven-plugin:run \
  -Drewrite.activeRecipes=org.openrewrite.java.RemoveUnusedImports
```

but sadly [only for recipes that require no configuration][:(]. Recipes that do require configuration have no (canonical) inline invocation form, instead necessitating the creation of a verbose and complex file merely to communicate what is often trivial information to the recipe, and then you even _still_ need the `rewrite.activeRecipes` user property from the example. This severely limits the practical utility of OpenRewrite. Instead of being a hypothetical arbitrary rename-refactor the tool is only really interesting for completely static transformations, or for dynamic transformations that are particularly complex or that have to run in batch mode.

[:(]: https://docs.openrewrite.org/running-recipes/running-rewrite-on-a-maven-project-without-modifying-the-build#running-a-recipe-with-configuration-parameters

## What problem are you trying to solve?

Consider a trivial recipe with a structure comparable to

```yaml
---
type: specs.openrewrite.org/v1beta/recipe
name: com.yourorg.ChangePackageExample
displayName: Rename package name example
recipeList:
  - org.openrewrite.java.ChangePackage:
      oldPackageName: com.yourorg.foo
      newPackageName: com.yourorg.bar
```

Suppose this needs to be run by _many different individuals_ for different values of `oldPackageName` or `newPackageName`. For \<reasons> we can distribute instructions for defining and executing a recipe but distributing a recipe itself is impractical; say, for example, that we cannot know literally every variation of package names, or that the old package name can exist in multiple artifacts and should be turned into numerous divergent package names.

The YAML configuration file works well with distribution when distribution is already a solved problem, but distribution is a difficult problem to solve and cumbersome even when already solved. It requires a distribution channel, which you may not already have, and if you do have one, using it may necessitate a disproportionate amount of bureaucracy namely if you're subject to auditing regulations. Even if distribution is feasible, some recipes are impossible to distribute in practice, like the examples from the previous paragraph.

But I don't see why we couldn't "just" pass the package names as user properties like we can with `rewrite.activeRecipes`.

## Describe the solution you'd like

Imagine I could run something like

```sh
mvn -U org.openrewrite.maven:rewrite-maven-plugin:run \
[...]

PR #816, opened 2024-07-08
Pass in single recipe configuration options on the command line

velo:
- Possible fix for https://github.com/openrewrite/rewrite-maven-plugin/issues/799

Use like this:

```
mvn org.openrewrite.maven:rewrite-maven-plugin:5.36.0-SNAPSHOT:run \
  -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-migrate-java:RELEASE \
  -Drewrite.activeRecipes=org.openrewrite.java.migrate.util.ReplaceStreamCollectWithToList  \
  -Drewrite.options=convertToList=true
```


timtebeek:
Thanks for getting this started! Would you happen to have an example of how  the non-flat Map is assigned from the command line? I'm wondering how to test this properly. We'll likely start with incubating this, and potentially have to limit it to non-composite recipes, as larger composites can contain a single recipe multiple times, and not all recipes might have the same options. Right now it seems like we might set the same argument on all recipes?

velo:
From command line it wouldn't be possible, as maven don't parse complex types from command line.

Those can only be set as maven configuration on pom file

```
<options>
  <convertToList>true</convertToList>
</options>
```

src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
```diff
@@ -126,6 +148,62 @@ protected ResultsContainer listResults(ExecutionContext ctx) throws MojoExecutio
...
+            CompositeRecipe composite = (CompositeRecipe) recipe;
+            for (Recipe child : composite.getRecipeList()) {
+                this.configureRecipeOptions(child);
+            }
+        } else if (recipe instanceof DeclarativeRecipe) {
+            DeclarativeRecipe declarativeRecipe = (DeclarativeRecipe) recipe;
+            for (Recipe child : declarativeRecipe.getRecipeList()) {
+                this.configureRecipeOptions(child);
+            }
+        } else {
```
timtebeek, line 162: From the discussion on the issue I think it's best that we don't try to support Declarative or Composite recipes just yet.
- https://github.com/openrewrite/rewrite-maven-plugin/issues/799

These would be hierarchical, possibly contain the same recipe multiple times, as well as other recipes that do not have the same options. As such they are highly likely to trigger exceptions when trying to `updateOption` below.

In terms of how to fail I suggest we validate early that the recipe is not a Composite/Declarative recipe when any options are set.

src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
```diff
@@ -54,6 +62,10 @@ public abstract class AbstractRewriteBaseRunMojo extends AbstractRewriteMojo {
 
     @Parameter(property = "rewrite.exportDatatables", defaultValue = "false")
     protected boolean exportDatatables;
+    @Parameter(property = "rewrite.options")
+    protected Map<String, String> options;
+    @Parameter(property = "rewrite.flatOptions")
+    protected String flatOptions;
```
timtebeek, line 68: I'm leaning towards only having one or the other, as we're mostly targeting one-off single recipe runs here, I think a flow aimed exclusively at command line argument handling is best. I wouldn't expect or suggest folks configure single recipes in their `pom.xml`, as the better fit there would be a `rewrite.yml` file containing those same options.

timtebeek (commented):
Nice to see this kicked off ; I've added two comments that I think we should address before a merge, but I'm also interested to hear your thoughts. Mostly I want to add this in such a way that we promote good practices, and not enable anti-patterns too much.

src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
velo: The more a use rewrite maven plugin, the more I think it needs to be slightly different...

For this, `flatOptions` is a must have for command line, as `options` don't work from command line due to maven not parsing complex types from command line argument.

But, I think `recipes` should be different, like this:


```
  <activeRecipes>
    <activeRecipe>
      <name>org.openrewrite.java.testing.hamcrest.MigrateHamcrestToAssertJ</name>
      <exclusions>
        <exclusion>**/pom.xml</exclusion>
      </exclusions>
    </activeRecipe>
    <activeRecipe>
      <name>org.openrewrite.java.testing.hamcrest.HamcrestMatcherToAssertJ</name>
      <options>
        <option>
          <matcher>fooBar</matcher>
          <assertion>isFoo</assertion>
          <argumentType>my.Bar</argumentType>
        </option>
      </options>
    </activeRecipe>
    <activeRecipe>
      <name>org.openrewrite.java.testing.assertj.JUnitToAssertj</name>
    </activeRecipe>
    <activeRecipe>
      <name>org.openrewrite.java.testing.assertj.Assertj</name>
    </activeRecipe>
    <activeRecipe>
      <name>org.openrewrite.java.migrate.UpgradeToJava21</name>
      <exclusions>
        <exclusion>**/src/main/java/**</exclusion>
        <exclusion>**/pom.xml</exclusion>
      </exclusions>
    </activeRecipe>
  </activeRecipes>
```

This would be very useful for complex automatically review PRs with OpenRewrite recipes.

src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
velo: Makes sense. I can undo this

src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
timtebeek: I'd hold off any such changes; we have other ways to support such a workflow if needed, for instance using profiles, or add your recipes separated by `---` to a `rewrite.yml` in the root of your project. That last option has the benefit of also being supported in IntelliJ Ultimate by default.

The more complicated configuration options seem like they would be challenging to maintain going forward. We'd really want to push the configuration into options supported by Maven, Gradle, Moderne CLI, JetBrains' OpenRewrite plugin for IntelliJ, Moderne Platform, ... Having a structure that works for all those options is then preferred over something unique to the Maven plugin.

timtebeek (approved):
I've gone ahead and applied the changes I suggested earlier @velo ; Hope you agree with the final form. I'll likely merge this tomorrow morning unless any blocking feedback comes in.

src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
```diff
@@ -55,6 +59,10 @@ public abstract class AbstractRewriteBaseRunMojo extends AbstractRewriteMojo {
     @Parameter(property = "rewrite.exportDatatables", defaultValue = "false")
     protected boolean exportDatatables;
 
+    @Parameter(property = "rewrite.options")
+    @Nullable
+    protected String options;
```
timtebeek, line 64: One thing left to try here is if this could just be a list to remove one of the split operations below

velo: Not for when using command line ARGS, maven don't parse complex types

timtebeek: Hmm; I thought we had that working for these two options by using LinkedHashSet
https://github.com/openrewrite/rewrite-maven-plugin/blob/1d2d8aa4ac246e7b0d47d3b4762c96379cb7ee92/src/main/java/org/openrewrite/maven/ConfigurableRewriteMojo.java#L54-L60

My tests are hampered by Sonatype OSS unavailability, so I can't test right now, but I'd like to give that a shot before a merge.

timtebeek:
This now works.
```bash
mvn org.openrewrite.maven:rewrite-maven-plugin:5.36.0-SNAPSHOT:run -Drewrite.activeRecipes=org.openrewrite.maven.RemovePlugin -Drewrite.options=groupId=org.springframework.boot,artifactId=spring-boot-maven-plugin
```

When run against the spring-petclinic it correctly removes only the argument plugin;
When a second recipe is activated, or a required option missing, we reject the recipe run.

Thanks to all involved!

velo:
This is awesome... I'm still shocked that maven can parse lists but can't maps

src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
```diff
@@ -126,6 +137,76 @@ protected ResultsContainer listResults(ExecutionContext ctx) throws MojoExecutio
         }
     }
 
+    private static void configureRecipeOptions(Recipe recipe, Set<String> options) throws MojoExecutionException {
+        if (recipe instanceof CompositeRecipe && recipe.getRecipeList().size() == 1) {
+            // Unpack active recipe if it's a single recipe
+            recipe = recipe.getRecipeList().get(0);
+        }
```
timtebeek, line 144: - No longer necessary after https://github.com/openrewrite/rewrite/pull/4349
```suggestion
```

timtebeek: This will be the final touch; then we can finally see this resolved and update this section of the docs:
https://github.com/openrewrite/rewrite-docs/blob/9f8f51d17b28248624487eecd15d352fc123ca2e/reference/faq.md?plain=1#L136-L143

timtebeek:
Thanks again @velo ! Hope this helps folks to do quick one-off configurable recipe runs.

PR #832, opened 2024-08-06
Drop CycloneDX goal after recipe addition

timtebeek:
We fairly aggressively try to only have one way to do things; this recipe is the new intended replacement. https://github.com/openrewrite/rewrite-java-dependencies/commit/90a31e7ad1a28cb2688023afe2dfb8e3c85058bc


timtebeek:
Dropping as discussed on https://www.youtube.com/watch?v=7KNj5bEZadE

Issue #851, opened 2024-08-30
Later build phases depending on earlier rewrite:runNoFork not supported in multi module Maven project


runeflobakk:
## What version of OpenRewrite are you using?

I am using

- OpenRewrite v8.34.0
- rewrite-maven-plugin v5.39.2
- rewrite-migrate-java v2.23.0

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a multi module project.

I want to rewrite source code in an early phase of the build (process-sources), in order to have the rewritten code participate in the usual compiling phase. This seems to work when building only the single module the way I would intuitively set it up: by binding `runNoFork` to a `process-sources` in my POM:

```xml
<plugin>
    <groupId>org.openrewrite.maven</groupId>
    <artifactId>rewrite-maven-plugin</artifactId>
    <executions>
        <execution>
            <id>transform-to-jakarta-sources</id>
            <phase>process-sources</phase>
            <goals>
                <goal>runNoFork</goal>
            </goals>
            <configuration>
                <activeRecipes>
                    <recipe>org.openrewrite.java.migrate.jakarta.JavaxXmlBindMigrationToJakartaXmlBind</recipe>
                </activeRecipes>
            </configuration>
        </execution>
    </executions>
    <dependencies>
        <dependency>
            <groupId>org.openrewrite.recipe</groupId>
            <artifactId>rewrite-migrate-java</artifactId>
            <version>2.23.0</version>
        </dependency>
    </dependencies>
</plugin>
```

Everything appears to work really well when working with the Maven module in question, I get the source files rewritten as they should, but when running the entire multiproject, I get the following logged at the stage where I have configured rewrite-maven-plugin to execute:
```xml
[INFO] --- rewrite:5.39.2:runNoFork (transform-to-jakarta-sources) @ my-rewriting-maven-module ---
[INFO] REWRITE: Delaying execution to the end of multi-module project for group.id:my-rewriting-maven-module:1.0-SNAPSHOT
```
And then further goals in the build fail because they depend on the code being rewritten.

I suspect my use-case is very similar to #755 , though I do not depend on the build-helper-maven-plugin. I use maven-dependency-plugin:unpack on an existing sources artifact, unpacking to `src/main/java`, and then wanted to run `runNoFork` on the unpacked sources for them to participate in the rest of the build, like ordinary Java source files.

[...]

Issue #755, opened 2024-03-14
Rewrite sources added through `build-helper-maven-plugin`


Rapster:
Based on https://stackoverflow.com/q/77927116/4605161

## What problem are you trying to solve?
In our codebase, we had to copy an entire class by hand just because some methods are private. So instead of being able to overwrite a few methods, we end up maintining a class of thousands LOC.

## Describe the solution you'd like
After copying the entire source code with `build-helper-plugin`, I'd love to apply a few transformations with recipes (essentially private to protected). Execute the recipe maven plugin in process-sources phase to execute a recipe

## Have you considered any alternatives or workarounds?
Spoon

timtebeek:
Thanks for logging an issue @Rapster ! We'd up to now been tracking this internally, but good to have a public facing counterpart as well for the wider community to see, track progress and potentially help out.

I haven't looked at the specifics yet, but it's likely to involve changes to our MavenMojoProjectParser, likely around here.
https://github.com/openrewrite/rewrite-maven-plugin/blob/41fa9a252d12fb477a0122f4c2a5dd63bac17d9a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L182-L183

You'll notice we have access to both a `MavenProject` and `Xml.Document maven` there, which should I think give access to any configured build-helper-maven-plugin in the same project. From there it's likely a case of parse those additional sources with the correct classpath, and we should then be set to make code changes.

Any help appreciated!

Rapster:
To realize what I want to do here, I need the goal to be executed in `process-sources` phase, is it feasible? AFAICS, rewrite goals are bound to `process-test-sources`

timtebeek:
> To realize what I want to do here, I need the goal to be executed in `process-sources` phase, is it feasible? AFAICS, rewrite goals are bound to `process-test-sources`

Not sure if that's feasible as a single build step, as indeed we tie to `process-test-classes`;
https://github.com/openrewrite/rewrite-maven-plugin/blob/41fa9a252d12fb477a0122f4c2a5dd63bac17d9a/src/main/java/org/openrewrite/maven/RewriteRunMojo.java#L29-L32

Would having to run multiple commands rule out your use case?

Rapster:
I guess it should be fine, depends what those steps would be. But maybe instead of relying on maven, I could do it programmatically (using rewrite API) but can't much doc there. Any idea?

timtebeek:
[...]

PR #852, opened 2024-08-31
Support configure log level for recipe changes

runeflobakk:
## What's changed?

Adding configurable log level for when running recipe(s) result in changes:
```xml
<configuration>
    <recipeChangeLogLevel>{DEBUG|INFO|WARN|ERROR}</recipeChangeLogLevel>
```

The default is "WARN" to align with existing behaviour.
It only affects logging from recipe results (which are currently hardcoded to `.warn(..)`). Any other logging is intentionally not affected by this property.

## What's your motivation?

I, and I see indications from others as well, use OpenRewrite to _always_ rewrite existing code from some source. I.e. the rewriting and changes are expected and handled by further steps in a build. For these use cases it is not appropriate to log performed changes as warnings. The motivation is to keep a clean build log, and only issue warnings for actual issues.

Example use-cases I believe are relevant for this change: 
- #755, 
- #851

The name of the configuration property. I currently named is as `recipeChangeLogLevel`, with the intention to communicate that it controls the log level used for logging changes performed by recipes. I am very open for any better name, if anyone has an idea for it 😅 

### Checklist

I am one of those people coding in Eclipse, but I believe the small changes introduced here should not be crashing with how IntelliJ would format the source files.


timtebeek (commented):
Great to see! Indeed seems like a valid use case; I've minimized the changes & API by folding `log` under `AbstractRewriteBaseRunMojo`, as opposed to a separate class that requires repeated arguments.

timtebeek (approved):

src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
```diff
@@ -62,6 +62,29 @@ public abstract class AbstractRewriteBaseRunMojo extends AbstractRewriteMojo {
     @Nullable
     protected LinkedHashSet<String> options;
 
+    /**
+     * The level used to log changes performed by recipes.
+     */
+    @Parameter(property = "rewrite.recipeChangeLogLevel", defaultValue = "WARN")
+    protected LogLevel recipeChangeLogLevel;
+
+    protected void log(CharSequence content) {
```
runeflobakk, line 71: All fine with this change. Just wanted to explain why I implemented it the way I originally did.

Now you essentially have a method in an abstract class, available in all the different implementations, and it only communicates "log" and a string to be logged. It is very implicit that the actual log level is affected by a LogLevel field with very specific semantics. I didn't want to introduce a method named only "log", which is intended for a very specific use: to log something that changed due to a rewrite.

To me, it seems better to have a method which accepts an explicit log level (e.g. the configured recipeChangeLogLevel), because the knowledge about what is logged is at the site where one invokes the logging-method. Now one really has to _know_ that the method named only "log" is specifically for logging changes made from a recipe, instead of the method accepting the loglevel as a parameter and carrying out as instructed, and the knowledge about why a certain loglevel is used is kept at the call site.

This can be implemented as a method accepting the LogLevel in the abstract class, absolutely, and that would also align with my original "design". Externalizing it to a separate (package scoped) class is just my preference to not tie too much concerns into abstract classes and making them available through inheritance.

Just wanted to mention it for you to consider :)

runeflobakk: A (better IMHO) alternative would be to perhaps have
```java
protected void log(LogLevel level, CharSequence content) {
```
instead, as it would not tie the log-method implicitly to the `recipeChangeLogLevel`. And you avoid the separate class.

It would be very easy for other code to evolve to invoke the new log-method accepting only the message, and not logging anything related to "changes from rewriting", but gets the log level unintentionally configured from `recipeChangeLogLevel`.

runeflobakk: If you prefer not repeating (passing) `recipeChangeLogLevel` from the different logging sites, I would rename the logging-method specifically for its intended use, perhaps `logRecipeChange(CharSequence)` or something, to communicate its intended use. It is not a _general_ log method.

timtebeek: Thanks for that context! I've adopted your proposal of `log(LogLevel, CharSequence)`

runeflobakk (commented):
👍👍👍
Thanks for a very swift review! :)

PR #868, opened 2024-09-23
Accept any directories under `src` in `ResourceParser`

marcel-gepardec:
Added accepted directories to accept all recursive directories within "src".

## What's changed?
Previously it only checked whether the path contains one of the DEFAULT_IGNORED_DIRECTORIES. 
Now it checks whether the path also contains one of the DEFAULT_ACCEPTED_DIRECTORIES.
Now "src" will return false almost instantly and will not return true although there could be a build or out folder within the "src" folder.

## What's your motivation?
Motivation was the open Issue #850. This is a fix to this issue and maybe a feature to be considered.

- Fixes https://github.com/openrewrite/rewrite-maven-plugin/issues/850

### DIFF
--- a/src/main/java/org/openrewrite/maven/ResourceParser.java
+++ b/src/main/java/org/openrewrite/maven/ResourceParser.java
@@ -40,10 +40,10 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class ResourceParser {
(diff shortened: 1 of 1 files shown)


timtebeek (approved):
Nice and measured improvement here @marcel-gepardec , thanks!

PR #882, opened 2024-10-24
Fix download snapshot warnings

ammachado:
## What's changed?
Fix download warnings for SNAPSHOT components

## What's your motivation?
- Fix https://github.com/openrewrite/rewrite/issues/4508

No

@timtebeek 

No

## Any additional context
No.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -611,6 +611,7 @@ public class MavenMojoProjectParser {
      *
      * @param project A maven project to examine for any children/parent poms.
      * @param paths   A list of paths to poms that have been collected so far.
+     * @param ctx     The execution context for the current project.
      */
     private void collectPoms(MavenProject project, Set<Path> paths, MavenExecutionContextView ctx) {
         if (!paths.add(pomPath(project))) {
@@ -618,7 +619,7 @@ public class MavenMojoProjectParser {
         }
 
         ResolvedGroupArtifactVersion gav = createResolvedGAV(project, ctx);
-        ctx.getPomCache().putPom(gav, createPom(gav, project));
+        ctx.getPomCache().putPom(gav, createPom(project));
 
         // children
         if (project.getCollectedProjects() != null) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -800,4 +804,22 @@ private void logDebug(MavenProject mavenProject, String message) {
...
+                project.getVersion().endsWith("-SNAPSHOT") ? null : project.getVersion()
+        );
+    }
+
+    private static Pom createPom(ResolvedGroupArtifactVersion gav, MavenProject project) {
+        Pom.PomBuilder builder = Pom.builder()
+                .gav(gav)
+                ;
+        // TODO: complete mapping
+        return builder.build();
```
timtebeek, line 823: I guess this is why it's still a WIP; let me know your plans and whether you'd need any help.

timtebeek (commented):
Great start! Thanks for the various improvements throughout and the cache introduction to fix the underlying issue. Let me know once you've worked through the final todo before a last review & merge.

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -807,4 +816,23 @@ private void logDebug(MavenProject mavenProject, String message) {
...
+                project.getArtifactId(),
+                project.getVersion(),
+                project.getVersion().endsWith("-SNAPSHOT") ? null : project.getVersion()
+        );
+    }
+
+    private static @Nullable Pom createPom(ResolvedGroupArtifactVersion gav, MavenProject project) {
+        try (InputStream is = Files.newInputStream(project.getFile().toPath())) {
+            RawPom rawPom = RawPom.parse(is, null);
+            return rawPom.toPom(project.getBasedir().toPath(), null);
```
ammachado, line 833: Should we use `ctx.getLocalRepository().getUri()` here too?

timtebeek (commented):
Great to see this @ammachado ; it was a bit of a puzzle for me correlating this with the MavenParser in openrewrite/rewrite, but I came up with a couple of questions that would be good to address I think. Curious to hear your thoughts, as you've perhaps already looked into those aspects.

I've also added a minimal test since I noticed none were using snapshot versions before. We might want to extend that with some assertions, although that might be a bit of a challenge with the testing framework used for plugins.

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
ammachado: I copied the idea from `MavenParser`.

timtebeek (approved):
Thanks again @ammachado ! Great to see these cleared out.

PR #888, opened 2024-10-31
Add tests for multi source sets

nielsdebruin:
## What's changed?
This PR adds support and test cases for running recipes on **test** source files located outside the standard maven test sources directory `src/test/java` for example `src/integration-test/java`. 

**Important**: The PR showed a bug with our current integration testing method for the Maven plugin. This causes the test cases that have been added to fail, even though manual testing of the plugin by myself and @timtebeek showed it works as desired. For this reason, the added tests have currently been disabled. 

## What's your motivation?
Currently, recipes that do not rely on classpath resources, e.g., rewriting a boolean expression, will still work. However, recipes that rely on classpath resources such as `FindTypes` will not work.

- fixes https://github.com/openrewrite/rewrite-maven-plugin/issues/755

### Checklist


src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -458,7 +458,10 @@ private Stream<SourceFile> processTestSources(
         javaParserBuilder.classpath(testDependencies).typeCache(typeCache);
         kotlinParserBuilder.classpath(testDependencies).typeCache(new JavaTypeCache());
 
-        List<Path> testJavaSources = listJavaSources(mavenProject.getBasedir().toPath().resolve(mavenProject.getBuild().getTestSourceDirectory()));
+        Set<Path> testJavaSources = new LinkedHashSet<>(listJavaSources(mavenProject.getBasedir().toPath().resolve(mavenProject.getBuild().getTestSourceDirectory())));
```
Laurens-W, line 461: Why wrap it in a set? :)

timtebeek:
I noticed that in Maven Core the build source directories are added to the project source roots as well;
https://github.com/apache/maven/blob/f8cba5dc2717230a17c9589efe111d8dd0c1abbf/impl/maven-core/src/main/java/org/apache/maven/project/DefaultProjectBuilder.java#L576-L582

Are we sure we need to separately add those build source directories here? Or can we rely on the above?

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
nielsdebruin: It is testJavaSource used later in the function, after this change it can contain duplicates, which will break things.

timtebeek: No longer needed after [clearing out duplicates](https://github.com/openrewrite/rewrite-maven-plugin/pull/888/commits/093f3456e72ef33fceb7f904b1262b338c88b530).

timtebeek (approved):

PR #923, opened 2025-01-03
Updating Suppressions

lkerford:
removing unnecessary suppressions
updating the until date for suppressions which still apply


timtebeek:
Were you aware of this recipe we have? https://app.moderne.io/recipes/org.openrewrite.java.dependencies.RemoveExpiredSuppressions

We could also add another one that removes suppressions for transitive dependencies that no longer come in transitively.

timtebeek (approved):

lkerford:
> Were you aware of this recipe we have? https://app.moderne.io/recipes/org.openrewrite.java.dependencies.RemoveExpiredSuppressions
> 
> We could also add another one that removes suppressions for transitive dependencies that no longer come in transitively.

no I didn't, thanks for sharing this is going to be very useful

Issue #940, opened 2025-02-01
Some errors logged only with Maven DEBUG log level


philippe-granet:
## What problem are you trying to solve?
Some errors are logged with WARN log level:
```
[INFO] Project [openrewrite-demo] Resolving Poms...
[INFO] Project [openrewrite-demo] Parsing source files
[WARNING] There were problems parsing some source files
[WARNING] There were problems parsing openrewrite-demo-test/manifest.yaml
[WARNING] org.yaml.snakeyaml.scanner.ScannerException: while scanning for the next token
found character '%' that cannot start any token. (Do not use % for indentation)
 in 'reader', line 15, column 6:
        {% ...
         ^

  org.yaml.snakeyaml.scanner.ScannerImpl.fetchMoreTokens(ScannerImpl.java:445)
  org.yaml.snakeyaml.scanner.ScannerImpl.checkToken(ScannerImpl.java:238)
  org.yaml.snakeyaml.parser.ParserImpl$ParseBlockMappingKey.produce(ParserImpl.java:637)
  org.yaml.snakeyaml.parser.ParserImpl.peekEvent(ParserImpl.java:161)
  org.yaml.snakeyaml.parser.ParserImpl.getEvent(ParserImpl.java:170)
```
but some errors are logged only if I activate DEBUG on Maven CLI (`--debug`):
```
[INFO] Running recipe(s)...
[DEBUG]
java.lang.IllegalArgumentException: Illegal group reference
    at java.util.regex.Matcher.appendExpandedReplacement (Matcher.java:1113)
    at java.util.regex.Matcher.appendReplacement (Matcher.java:1043)
    at java.util.regex.Matcher.replaceAll (Matcher.java:1229)
    at org.openrewrite.text.FindAndReplace$1.visit (FindAndReplace.java:150)
    at org.openrewrite.text.FindAndReplace$1.visit (FindAndReplace.java:110)
    at org.openrewrite.TreeVisitor.visit (TreeVisitor.java:157)
    at org.openrewrite.Preconditions$Check.visit (Preconditions.java:175)
    at org.openrewrite.Preconditions$Check.visit (Preconditions.java:145)
```

## Describe the solution you'd like
Can choose logging level for all errors, and verbosity (with/without stacktrace)

timtebeek:
[...]

PR #942, opened 2025-02-04
Report first `Throwable` caught in `InMemoryExecutionContext.onError`

timtebeek:
## What's your motivation?
Makes it easier to troubleshoot recipe execution errors that do not result in source related exceptions.

- Fixes https://github.com/openrewrite/rewrite-maven-plugin/issues/940

Now we log a warning for the first caught throwable, but do not halt execution or report subsequent warnings.

@philippe-granet 

We could choose to log warnings beyond the first, or even halt execution. That's a more drastic change though, and perhaps good to first, for a while, raise the visibility of these caught throwables.

## Any additional context
- https://github.com/openrewrite/rewrite-maven-plugin/issues/940#issuecomment-2635127698

### DIFF
--- a/src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
+++ b/src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
@@ -47,6 +47,7 @@ import java.time.Duration;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.*;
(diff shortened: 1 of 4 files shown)


philippe-granet (commented):
Looks great for me. Perhaps you can add a message when option exportDatatables is disabled. This will indicate users they can retrieve all errors by this way

timtebeek:
Thanks again for the suggestion! Should help uncover such issues going forward.

PR #949, opened 2025-02-17
fix: set TCCL to recipe classloader for SPI resolution

MBoegers:
## What's your motivation?

Set TCCL explictely to Recipes CL to be enbale SPI to load classes from within the Recipe Artifact

### Checklist


MBoegers:
seems to be not needed.

PR #967, opened 2025-03-24
Check recipe License

MBoegers:
## What's changed?

Prevent OS Plugin users to accidentally execute proprietary recipes.
Test have to be refined.

### Checklist

### DIFF
--- a/src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
+++ b/src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
@@ -26,6 +26,7 @@ import org.openrewrite.*;
 import org.openrewrite.config.CompositeRecipe;
 import org.openrewrite.config.DeclarativeRecipe;
 import org.openrewrite.config.Environment;
+import org.openrewrite.config.License;
 import org.openrewrite.config.RecipeDescriptor;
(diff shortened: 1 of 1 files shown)


src/main/java/org/openrewrite/maven/AbstractRewriteBaseRunMojo.java
```diff
@@ -148,6 +149,14 @@ protected ResultsContainer listResults(ExecutionContext ctx) throws MojoExecutio
                 }
             }
 
+            getLog().info("Check recipe licences...");
+            failedValidations = checkLicence(recipe);
```
timtebeek, line 153: Right now we're validating the active recipe and any recipes that contains. I had expected we'd already throw earlier when loading recipes at all. Now folks can still misconfigure and only find out depending on what they run. Do you see an option for that?

MBoegers: We could validate at classloading, but for the Maven plugin thats only seconds, or less, before validating the active recipe like now. 
Maybe we could hook into the resolution of the dependencies and break when a artifact is downloaded that is Moderne Prop. but again seconds before the current point.
So what advantage do you see and which goal you have in mind?

PR #969, opened 2025-03-28
Remove gauge warning

rlsanders4:
## What's changed?

Fix the Gaguge issue that was causing this warning:
```
[WARNING] This Gauge has been already registered (MeterId{name='cache.size', tags=[tag(cache=Maven POMs - default)]}), the Gauge registration will be ignored. Note that subsequent logs will be logged at debug level.
```

@timtebeek 

### Checklist


timtebeek:
Awesome, thanks a lot for locating and fixing this so quickly!

PR #978, opened 2025-04-23
feat: better plainTextMask for Containerfile, Dockerfile

jimschubert:
## What's changed?

* Improves Dockerfile plain text mask to support other patterns (`dockerfile`, `dockerfile.build`, `build.dockerfile`, `release.Dockerfile`, etc.)
* Adds Containerfile plain text mask (see https://github.com/openrewrite/rewrite-docker/issues/3)

### Docker's Documentation

For casing, [Docker's documentation](https://docs.docker.com/build/concepts/dockerfile/#filename) says:

> [Filename](https://docs.docker.com/build/concepts/dockerfile/#filename)
> 
> The default filename to use for a Dockerfile is `Dockerfile`, without a file extension. Using the default name allows you to run the docker build command without having to specify additional command flags.
> 
> Some projects may need distinct Dockerfiles for specific purposes. A common convention is to name these `<something>.Dockerfile`. You can specify the Dockerfile filename using the --file flag for the docker build command. Refer to the [docker build CLI reference](https://docs.docker.com/reference/cli/docker/buildx/build/#file) to learn about the --file flag.
> 
> Note
> 
> We recommend using the default (`Dockerfile`) for your project's primary Dockerfile.

### Docker Undocumented Behavior

They don't explicitly call it out, but `dockerfile` is allowed for the primary dockerfile and honored by `docker build` without the `--file` flag:

image

Other build systems such as podman will require the title cased file:

image

### Conventions in the Ecosystem

While Docker's documentation suggests `<something>.Dockerfile`, even official images don't follow that recommendation (example: https://github.com/docker-library/mysql/blob/8ade9b2c9a32a79fbaa44b564d09a40744f1d105/Dockerfile.debian).

### Containerfile alignment

As described in https://github.com/openrewrite/rewrite-docker/issues/3, `Containerfile` should also be supported. I'd like to use Containerfile without providing the additional plain text mask.

no

no

Providing the patterns manually for 3 of the 4 use cases, but that's weird.

## Any additional context

Not related to this implementation, but I have been working on [jimschubert/rewrite-docker](https://github.com/jimschubert/rewrite-docker) and it is coming along (although manually parsing docker is a bit cumbersome). This is a major factor in why I'm looking for the additional Dockerfile and Containerfile naming conventions to be supported out of the box by the maven plugin, as supporting only `Dockerfile*` is confusing.

Related to the checklist below: I don't know how I would apply a negative test case for the plain text matching which isn't applied elsewhere.

### Checklist


timtebeek (approved):
Perfect PR here @jimschubert , zero remarks other than thanks a lot! Will go out with the next release

PR #986, opened 2025-05-16
Create type tables for argument GAVs in Maven

timtebeek:
## What's changed?
Add a new goal to the rewrite-maven-plugin to create type tables.

## What's your motivation?
Parity for folks using Maven, as up to now that had only been [available for Gradle](https://github.com/openrewrite/rewrite-build-gradle-plugin/blob/17964990a5d33cc6c025c315bd87e5eb0e3fd0eb/src/main/java/org/openrewrite/gradle/RecipeDependenciesTypeTableTask.java#L32-L37).

Figured I'd reuse as much as possible, while working around the fact that Maven does not easily allow resolving two versions. By using a profile folks should be able to clearly document the steps to recreate a type table, without it interfering with recipe runs on the same recipe module.

@nmck257 I believe this was relevant for you all there right?

Briefly explored plugin `<dependency>`, but that would not allow for more than one version of an artifact, so figured use `recipeArtifactCoordinates` with GAV arguments instead.

## Any additional context
- https://github.com/openrewrite/rewrite-build-gradle-plugin/pull/75
- https://github.com/openrewrite/rewrite/pull/4932
- https://github.com/openrewrite/rewrite-docs/issues/350

### DIFF
--- a/src/main/java/org/openrewrite/maven/RewriteTypeTableMojo.java
+++ b/src/main/java/org/openrewrite/maven/RewriteTypeTableMojo.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 the original author or authors.
+ * Copyright 2020 the original author or authors.
  * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


sambsnyd (approved):

tkvangorder:
🎉

PR #1141, opened 2026-04-11
Validate DEVELOCITY_ACCESS_KEY format before build

timtebeek:
## Summary

The Develocity Maven extension requires `DEVELOCITY_ACCESS_KEY` in the format `server-host=access-key`, but the `gradle_enterprise_access_key` secret appears to be malformed (missing the server host prefix). This causes the extension to crash after every Maven invocation, failing both the outer build and all integration tests.

This adds a validation step that unsets the env var if it doesn't contain `=`, allowing CI to pass until the secret is corrected.

## Test plan
- [ ] Verify CI passes on this branch
- [ ] Update the `gradle_enterprise_access_key` secret to the correct `server-host=access-key` format for a permanent fix

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,12 @@ jobs:
           server-password: SONATYPE_PASSWORD
           gpg-private-key: ${{ secrets.OSSRH_SIGNING_KEY }}
           gpg-passphrase: SONATYPE_SIGNING_PASSWORD
+      - name: validate-develocity-access-key
+        run: |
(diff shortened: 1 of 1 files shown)


timtebeek:
Looks like this pipeline also broke; @greg-at-moderne would you happen to know how to set the right token here too?

timtebeek:
We'd also failed to notice, so I'm also adding this one:
 - https://github.com/openrewrite/rewrite-maven-plugin/pull/1140

Both of these we can merge already; it just that we wont have develocity until the token is fixed.

Issue #1107, opened 2026-02-08
Recipes no longer applied to generated sources after upgrading from 6.26 to  6.27+


baflQA:
## What version of OpenRewrite are you using?
6.26.0 updated to 6.27.0 and/or 6.28.1

- Maven plugin 6.28.1
- rewrite-_module_ v1.2.3

## How are you running OpenRewrite?

I am using the Maven plugin, and my project is a multi module project.
```xml
 <build>
        <plugins>
            <plugin>
                <groupId>org.openrewrite.maven</groupId>
                <artifactId>rewrite-maven-plugin</artifactId>
                <version>6.28.1</version>
                <configuration>
                    <activeRecipes>
                        <recipe>org.openrewrite.java.RemoveUnusedImports</recipe>
                        <recipe>org.openrewrite.java.OrderImports</recipe>
                    </activeRecipes>
                </configuration>
            </plugin>
        </plugins>
    </build>
```

## What is the smallest, simplest way to reproduce the problem?
I have a script which runs openapi-generator-cli to generate dtos from my api-specs. 
There are some imports added from templates, which are not required for each generated file. 
I use open-rewrite plugin to clean the imports up. 
After upgrading to 6.27.0 open rewrite maven plugin stopped removing the imports.
- Somehow I've got a feeling that the issue is with the rewrite core itself, maybe caused by https://github.com/openrewrite/rewrite/pull/6347

When I stage and commit generated files, and then apply the open rewrite recipes, the imports will be cleaned up again.

baflQA:
Yeah, I managed to find a workaround - committing the generated DTOs before running OpenRewrite does the trick. When there are no uncommitted changes, the imports get cleaned up properly.
That said, I was a bit surprised by this behavioral change between 6.26.0 and 6.28.1, especially since it's a minor version bump. It's quite a significant shift in how the tool works in practice. I'm wondering if this might be:

- Something that could benefit from being documented more explicitly?
- Or perhaps the behavior around uncommitted files is still being refined?

Either way, I think it would be really helpful if this kind of change was either clearly explained in the release notes/migration guides, or handled more consistently by the plugin itself. It would save folks from having to adjust their workflows unexpectedly.
Really appreciate all the work you're putting into this plugin - it's been a huge help for our project! Just thought it was worth bringing this up so maybe the documentation or behavior could be clarified for future releases.

santunioni:
[...]

Issue #968, opened 2025-03-26
Scan the generated sources but not apply change


addozhang:
## What problem are you trying to solve?
Here we have projects that use OAS to define API and generate codes. These generated codes locate at target/generated-sources.

During recipes running, these generated codes will be referred by others sources or used in scanning  phase.

## Describe the solution you'd like

Scan the generated sources but not apply changes to these generated ones.

## Have you considered any alternatives or workarounds?

We may can involve a new Marker like Immutable, and add Marker to generated sources. 

During changes apply, skip the sources with this Marker.

## Are you interested in [contributing this feature to OpenRewrite](https://github.com/openrewrite/.github/blob/main/CONTRIBUTING.md#new-features)?

Yes, if the feature accepted.

timtebeek:
Hi @addozhang ; Interesting suggestion! If I understand correctly you'd want the generated source to participate in the scanning phase of any ScanningRecipe, but not be visited in the edit phase? Or have their suggested changes discarded?

Right now that would not be possible, as for instance the Maven plugin already filters out generated source LSTs in the parser:
https://github.com/openrewrite/rewrite-maven-plugin/blob/b75abb0fcc3d7584b15ea1d2322a950ca53ed396/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L443-L448

Could you describe why you'd want to see this change in a little more detail? You've for now mostly described what you'd like to see, but for a change like this the motivation is important before we'd make such structural changes.

addozhang:
Yes, I see from the source that the plugin reads the generated code but eventually filters it out.

I now want to migrate our project from a commercial framework to springboot. These APIs use OpenAPI to describe the interface that we use OAS Generator to generate controllers for endpoints. Then use openrewrite to generate the implementation of the Controller. Those business logic are described using xml. With xmlst, we implement the logical code in the Controller or Service layer.

I hope my description is clear enough.

timtebeek:
That helps a bit, thanks. Do I understand correctly you'd want to migrate the generated classes, and then from that point onwards maintain them and stop relying on the code generator? I wonder if the code generator can be made to produce alternative outputs that already match your desired outcome, and if it makes sense to move your generated sources to `src/main` such that no plugin/framework changes are needed.

addozhang:
It might be a good idea to have the OAS Generator plug-in generate classes
to the src/main directory. The generated code output path is currently
configured in the parent pom, and I need to see if it can be configured.
[...]

PR #1142, opened 2026-04-11
Skip parsing generated sources; use compiled classes instead

### LINKED ISSUE #667: Do not parse generated sources as source files, but only as compiled classes
As identified on
- https://github.com/openrewrite/rewrite-maven-plugin/pull/665#discussion_r1405340228

I think it boils down to the question if `getCompileClasspathElements()` returns the classes compiled fom generated sources as well, which could depend on a number of different factors depending on how the code generation is configured.
https://github.com/openrewrite/rewrite-maven-plugin/blob/0b8fc330f355a96b6e5b7a103601c07f775d9de3/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L317-L324

Right now we define generated sources as anything we find in the `target/` directory, and parse those as source.
https://github.com/openrewrite/rewrite-maven-plugin/blob/0b8fc330f355a96b6e5b7a103601c07f775d9de3/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L302-L308

Before we then later filter out those generated sources.
https://github.com/openrewrite/rewrite-maven-plugin/blob/9a9428fbd211b79cebdec18a68b830e16434ade3/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java#L341-L346

It might be interesting to not parse those generated sources as source files but rather only as classpath entries. I'm not entirely sure that would work in all cases, but it could speed up builds for projects that heavily use generated sources.

timtebeek:
## Summary

- Generated sources under the build directory (e.g. `target/generated-sources/`) were being fully parsed into LSTs only to be filtered out afterward
- Since `getCompileClasspathElements()` already includes `target/classes` with the compiled `.class` files, type attribution works without parsing the generated sources
- Adds `filterGeneratedSourceRoots()` to exclude source roots under the build directory before scanning, applied consistently in `processMainSources()`, `processTestSources()`, and the `parsedPaths` pre-population
- This improves performance for projects that heavily use code generation (protobuf, MapStruct, annotation processors, etc.)

- Closes #667

## Impact on related issues

- **#968** — Requests generated sources be scanned (for search recipes) but not modified. This PR does not regress that use case: generated sources were already fully discarded after parsing, so search results from them were already lost. Properly supporting #968 would require a `GeneratedSource` marker, which is orthogonal to this change.
- **#1107** — Recipes no longer applied to generated sources after 6.27+. This PR does not change the filtering behavior introduced in 6.27 (PR #665); it only moves the exclusion earlier (before parsing instead of after). Users who want recipes applied to generated sources need a separate opt-in mechanism.
- **#649** — Maven Polyglot `ParseError`. Unrelated; that issue is about non-standard POM files, not source file handling.

## Test plan
- [x] Existing unit tests pass (16/16)
- [ ] Integration test with a project using annotation processing (e.g. MapStruct or Lombok) to verify type attribution still works
- [ ] Integration test with a protobuf project to verify generated types resolve correctly

### DIFF
--- a/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
+++ b/src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
@@ -238,14 +238,15 @@ public class MavenMojoProjectParser {
         Collection<PathMatcher> exclusionMatchers = exclusions.stream()
                 .map(pattern -> baseDir.getFileSystem().getPathMatcher("glob:" + pattern))
                 .collect(toList());
+        Path buildDirectory = baseDir.relativize(Paths.get(mavenProject.getBuild().getDirectory()));
         sourceFiles = sourceFiles.map(sourceFile -> {
-            if (isExcluded(repository, exclusionMatchers, sourceFile.getSourcePath())) {
+            if (isExcluded(repository, exclusionMatchers, buildDirectory, sourceFile.getSourcePath())) {
                 return null;
             }
             return sourceFile;
         }).filter(Objects::nonNull);
 
-        Stream<SourceFile> mavenWrapperFiles = parseMavenWrapperFiles(mavenProject, exclusionMatchers, parsedPaths, ctx);
+        Stream<SourceFile> mavenWrapperFiles = parseMavenWrapperFiles(mavenProject, exclusionMatchers, buildDirectory, parsedPaths, ctx);
         sourceFiles = Stream.concat(sourceFiles, mavenWrapperFiles);
 
         Stream<SourceFile> nonProjectResources = parseNonProjectResources(mavenProject, parsedPaths, ctx);
@@ -256,7 +257,10 @@ public class MavenMojoProjectParser {
                 .map(this::logParseErrors);
     }
 
-    static boolean isExcluded(org.openrewrite.jgit.lib.@Nullable Repository repository, Collection<PathMatcher> exclusionMatchers, Path path) {
+    static boolean isExcluded(org.openrewrite.jgit.lib.@Nullable Repository repository, Collection<PathMatcher> exclusionMatchers, Path buildDirectory, Path path) {
+        if (path.startsWith(buildDirectory)) {
+            return true;
+        }
         for (PathMatcher excluded : exclusionMatchers) {
             if (excluded.matches(path)) {
                 return true;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
```diff
@@ -222,12 +222,14 @@ public Stream<SourceFile> listSourceFiles(MavenProject mavenProject, Xml.@Nullab
 
         // Pre-populate parsedPaths with all source paths from both scopes
         // to prevent resource parsers from claiming cross-scope sources as PlainText
-        parsedPaths.addAll(listJavaSources(mavenProject, mavenProject.getExecutionProject().getCompileSourceRoots()));
+        List<String> mainSourceRoots = filterGeneratedSourceRoots(mavenProject, mavenProject.getExecutionProject().getCompileSourceRoots());
+        List<String> testSourceRoots = filterGeneratedSourceRoots(mavenProject, mavenProject.getExecutionProject().getTestCompileSourceRoots());
```
shanman190: Won't this result in the files being present in the model as PlainText? Maybe we should -- like the Gradle plugin -- filter out at the end? Avoiding parsing them ahead of time is useful as well though.

timtebeek: You're right, thanks! Changed the approach slightly to still claim the sources through `parsedPaths`, but not parse them in `processMainSources` and `processTestSources`. That should then still achieve the goals set out on #667 for efficiency, without affecting functionality.

shanman190 (commented):
Should we just add this to the isExcluded check on L241? That could probably be converted to a filter like we did in the Gradle plugin.

src/test/java/org/openrewrite/maven/MavenMojoProjectParserIsExcludedTest.java
```diff
@@ -168,6 +168,7 @@ void exclusionMatcherMatchesSubdirFileWithoutPrefixing() {
                 .isTrue();
     }
 
+
```
shanman190:
```suggestion
```

shanman190 (approved):

src/main/java/org/openrewrite/maven/MavenMojoProjectParser.java
shanman190: optional:  We could simplify this to a `.filter(sourceFile -> !isExcluded(repository, exclusionMatchers, sourceFile.getSourcePath()) && !source.getSourcePath().startsWith(buildDirectory))`

PR #1183, opened 2026-07-30
Allow resolving dependencies from the Code Genome Project

timtebeek:
Adds `https://artifacts.codegenomeproject.org/maven` as an additional source for `org.openrewrite` artifacts and forwards `CGP_ARTIFACTS_MAVEN_USERNAME` / `CGP_ARTIFACTS_MAVEN_TOKEN` to the CI build, the Maven counterpart to openrewrite/rewrite-github-actions@56b1b07 (where the Gradle convention plugin already does this). It sits alongside the existing repositories rather than replacing them, because CGP does not yet publish every snapshot we consume — `org.openrewrite:rewrite-bom:8.89.0-SNAPSHOT` returns 422 "no snapshot matching ... is published", while `rewrite-core`, `rewrite-java`, `rewrite-maven` and `rewrite-polyglot` are all present.

The repository is gated behind a `codegenome` profile activated by `CODEGENOME_USERNAME`, since CGP rejects anonymous requests and an unconditional declaration would 401 for contributors and fork PRs, which receive no secrets. `setup-java` is invoked a second time to write the default `~/.m2/settings.xml` the build step uses, as it emits only one server per call and the existing settings are read solely by the steps passing `--settings`. Publishing is untouched and still goes to Sonatype; switching that to CGP is a separate follow-up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,16 @@ jobs:
           server-password: SONATYPE_PASSWORD
           gpg-private-key: ${{ secrets.OSSRH_SIGNING_KEY }}
           gpg-passphrase: SONATYPE_SIGNING_PASSWORD
+      # setup-java writes a single server per call, and the settings above are only read by the
+      # steps that pass --settings; this second call supplies the default ~/.m2/settings.xml that
+      # the build step uses to authenticate against codegenome.
+      - uses: actions/setup-java@v5
+        with:
+          distribution: temurin
+          java-version: 17
+          server-id: codegenome
+          server-username: CODEGENOME_USERNAME
+          server-password: CODEGENOME_TOKEN
       - name: validate-develocity-access-key
         run: |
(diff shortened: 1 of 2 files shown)


timtebeek:
CI on this branch is red, and the cause is narrow: the build itself resolves fine (it falls back to Central snapshots for the one artifact CGP lacks), but the 422 surfaces as a Maven *warning*, and seven ITs assert that nested builds produce no warnings:

```
Expecting empty but was: ["Could not transfer metadata org.openrewrite:rewrite-bom:8.89.0-SNAPSHOT/maven-metadata.xml
  from/to codegenome (https://artifacts.codegenomeproject.org/maven): status code: 422, Unprocessable Entity"]
```

`RewriteDiscoverIT` (5) and `BasicIT` (2). Everything else — including resolving the other `org.openrewrite` snapshots from CGP — works, which confirms the credentials and settings wiring are correct.

Likely upstream root cause: `rewrite-bom/build.gradle.kts` applies `org.openrewrite.build.publish` but not `org.openrewrite.build.publish-cgp`, and `RewritePublishPlugin` does not apply the CGP plugin itself, so the BOM never gets CGP as a publishing target while the modules do.

PR #1201, opened 2026-08-25
Resolve org.openrewrite artifacts in the ITs from the Code Genome Project

timtebeek:
- The Maven counterpart to openrewrite/rewrite-gradle-plugin#471.

## What was wrong

- `#1183` pointed the main build at the Code Genome Project, but the itf integration tests were left out. Each `@MavenTest` forks its own Maven build against a local repository seeded from `target/itf-repo`. That repository carries the plugin under test and its runtime dependencies — it does not carry the recipe artifacts the fixtures pin as plugin dependencies, so those were resolved from Maven Central at test time, along with their transitive `org.openrewrite` trees.

Twelve IT poms pull `org.openrewrite` artifacts this way (`rewrite-static-analysis`, `rewrite-migrate-java`, `rewrite-all`, `rewrite-cloud-suitability-analyzer`, `rewrite-java`), plus `RewriteDryRunIT#no_plugin_in_pom`, which passes `rewrite.recipeArtifactCoordinates` and has the plugin resolve `rewrite-testing-frameworks` through the project's own repositories.

## What this does

1. Adds the same credential-gated `codegenome` profile the main build uses to each of those poms, as both a `<repository>` (project dependencies and `recipeArtifactCoordinates`, which `ArtifactResolver` resolves from `getRemoteArtifactRepositories()`) and a `<pluginRepository>` (plugin dependencies). Profile-injected repositories are source-dominant, so the Code Genome Project is consulted ahead of Central. Fork builds get no `CODEGENOME_USERNAME`, so the profile stays inactive there and they keep resolving from Central.

2. Bumps the pinned recipe versions to releases the Code Genome Project actually carries. This is the part that makes step 1 do anything: the repository only holds artifacts published after the cutover, so the old pins would have 404'd there and silently fallen back to Central.

   | artifact | before | after | earliest on CGP |
   | --- | --- | --- | --- |
   | `rewrite-static-analysis` | 1.0.4, 1.15.0 | 2.41.1 | 2.35.0 |
   | `rewrite-migrate-java` | 3.31.0 | 3.42.1 | 3.35.0 |
   | `rewrite-all` | 1.3.4 | 1.28.1 | 1.25.4 |
   | `rewrite-testing-frameworks` | 2.0.3 | 3.44.0 | 3.36.0 |
   | `org.openrewrite:rewrite-java` | 8.0.0 | 8.90.4 | 8.83.0 |

   The recipes each fixture activates were checked against the new jars: `SimplifyBooleanExpression`, `SimplifyBooleanReturn`, `UnnecessaryParentheses`, `AssertTrueNullToAssertNull`, `UpgradeToJava17` and `org.openrewrite.kotlin.format.AutoFormat` (via `rewrite-all` → `rewrite-kotlin`) are all still present. `ThrowingRecipe` still compiles against 8.90.4 — `org.openrewrite.internal.lang.Nullable` is deprecated but has not been removed.

## One declaration per recipe artifact

Six fixtures pin `rewrite-static-analysis`, and two of them had drifted to a different version than the other four (1.15.0 vs 1.0.4) before the version bump above collapsed them. To keep it that way, the version now lives in a single `<rewrite-static-analysis.version>` property on this pom and is filtered into the fixtures as `@rewrite-static-analysis.version@`.

That works because `src/test/resources-its` is already declared with `<filtering>true</filtering>` and itf's `resources-its` goal delegates to the standard `MavenResourcesFiltering` component — the same mechanism that resolves `@project.version@` in these poms today, so any project property is available.

Dependabot will not bump this one: it only tracks properties it can trace to a declared dependency, and this is free-floating. Same manual cadence as before, just one site instead of six.

`rewrite-all`, `rewrite-migrate-java` and `rewrite-java` get the same treatment. Every `org.openrewrite` artifact version under `src/test/resources-its` now comes from a property on this pom, so the Code Genome Project's version floor can be checked in one place instead of hunted across fixtures.

`rewrite-testing-frameworks` is the one exception, and it is a mechanism limit rather than a choice. Its coordinate is a `@SystemProperty` annotation value in `RewriteDryRunIT`, not a fixture pom, and `src/test/java` is not a filtered resource directory. itf does interpolate, but `PropertiesFilter` is applied only to `@MavenGoal` values and its map holds only `project.groupId`, `project.artifactId` and `project.version` — so `@...@` filtering cannot reach it and neither can itf's own interpolation. The workable alternatives are a compile-time constant next to the existing ones in `MavenCLIExtra`, or generating a constants class from the property; both trade the literal for indirection across two files, so I left it inline rather than pick one unilaterally.

**`maven-compiler-plugin` was deliberately not centralized**, despite looking like the biggest win: four fixtures pin it at three different versions (3.10.1, 3.13.0, 3.14.0). Two of those pins look load-bearing — `java_compiler_plugin_project` pairs 3.10.1 with an unresolvable `<release>${maven.compiler.release}</release>`, and `lombok_jdk25_linkage_error` pins 3.14.0 for a JDK 25 linkage error. Forcing them onto one version would quietly change what those fixtures cover, which is the opposite of what standardizing is for. Same reasoning for the JUnit 5.9.1 pin in `no_plugin_in_pom`, which feeds the `rewrite-testing-frameworks` recipe under test.

Remaining same-value repeats are all two-site and incidental: `kotlin.version` 1.9.10 and `jdk.version` 17 (both KotlinIT twins), `build-helper-maven-plugin` 3.6.0, `junit-jupiter` 5.9.1. No drift to prevent; happy to extract them if uniformity is worth the indirection.

## Replacing the cloud suitability fixture

`RewriteRunIT#cloud_suitability_project` ran `org.openrewrite.cloudsuitability.FindJks` over an empty `src/main/resource/some.jks`. What it actually covers is #517: `AbstractRewriteRunMojo` must report a changed `Quark` without trying to write to it.

`org.openrewrite.FindSourceFiles` reaches the same code path — its visitor marks any `SourceFile` whose path matches the glob, `Quark` included (it even special-cases `Quark` when filling in the data table's charset column), and returns `SearchResult.found(...)`. It lives in `rewrite-core`, which the plugin already ships, so the fixture now needs no external recipe artifact from any repository.

The fixture moves to `quark_project` with a `rewrite.yml` supplying `filePattern: '**/*.jks'`, matching the `container_masks` / `plaintext_masks` pattern for recipes that take options. The assertion is unchanged. Happy to keep the old directory name if the rename is unwelcome.

## Caveats worth a reviewer's attention

- **`rewrite-cloud-suitability-analyzer` was dropped rather than moved.** It is not on the Code Genome Project at all — `maven-metadata.xml` 404s, no versions — so it would have been the one `org.openrewrite` dependency still coming from Central. See the second commit.
- **Snapshots are disabled** on the Code Genome repository in these poms. Everything the fixtures need is a release, and enabling snapshots would have the inner builds fetch `maven-metadata.xml` for the plugin's own `-SNAPSHOT` version, which is the CDN-cached-metadata-vs-live-checksum mismatch that broke seven ITs on 2026-07-31. Many of these ITs assert their inner build emits no warnings.
- **Extra requests.** Adding a repository ahead of Central means every artifact the inner builds resolve — Maven core plugins, JUnit, Kotlin — is looked up on the Code Genome Project first and 404s before falling through. itf gives each test method a fresh local repository, so this repeats per test. Worth weighing against the alternative of pre-seeding `target/itf-repo`, which the conflicting pinned versions of `rewrite-static-analysis` make awkward.
- **What CI proves and what it does not.** CI is green: 38 tests run, 2 skipped, 0 failures, so the bumped recipe versions resolve and every fixture's assertions still hold. It does not by itself prove the artifacts came from the Code Genome Project rather than falling through to Central — `.mvn/maven.config` sets `--no-transfer-progress`, so neither the outer nor the inner builds log where anything was downloaded from. The load-bearing assumption is that `CODEGENOME_USERNAME` reaches the itf-forked builds and activates the profile there; that is ordinary env inheritance through the failsafe fork, and the 2026-07-31 checksum incident already showed Code Genome traffic reaching itf-forked builds. If we want it observed rather than inferred, the check is to inspect `_remote.repositories` under `target/maven-it/**/.m2/repository/org/openrewrite/` after a run. I could not do that locally — `repo.maven.apache.org` is returning 429 to Maven's HTTP client from this machine (curl gets 200; it is user-agent throttling), so `mvn verify` never got past resolving the Develocity extension.

No CI changes needed: `CODEGENOME_USERNAME` / `CODEGENOME_TOKEN` are already exported for the whole build step and inherited by the forked builds, and `~/.m2/settings.xml` already carries the `codegenome` server.


MBoegers (approved):
LGTM, only expected changes

